PR #4, opened 2016-09-16
add unit test for ValidateLayout function

xiekeyang:
migrate patch from https://github.com/opencontainers/image-spec/pull/310
cc @runcom @vbatts @wking @stevvooe

### DIFF
--- a/image/image_test.go
+++ b/image/image_test.go
@@ -1,0 +1,331 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const (
+	refTag = "latest"
+
+	layoutStr = `{"imageLayoutVersion": "1.0.0"}`
+
+	configStr = `{
+    "created": "2015-10-31T22:22:56.015925234Z",
+    "author": "Alyssa P. Hacker <alyspdev@example.com>",
+    "architecture": "amd64",
+    "os": "linux",
+    "config": {
+        "User": "alice",
+        "Memory": 2048,
+        "MemorySwap": 4096,
+        "CpuShares": 8,
+        "ExposedPorts": {
+            "8080/tcp": {}
+        },
+        "Env": [
+            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+            "FOO=oci_is_a",
+            "BAR=well_written_spec"
+        ],
+        "Entrypoint": [
+            "/bin/my-app-binary"
+        ],
+        "Cmd": [
+            "--foreground",
+            "--config",
+            "/etc/my-app.d/default.cfg"
+        ],
+        "Volumes": {
+            "/var/job-result-data": {},
+            "/var/log/my-app-logs": {}
+        },
+        "WorkingDir": "/home/alice"
+    },
+    "rootfs": {
+      "diff_ids": [
+        "sha256:c6f988f4874bb0add23a778f753c65efe992244e148a1d2ec2a8b664fb66bbd1",
+        "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
+      ],
+      "type": "layers"
+    },
+    "history": [
+      {
+        "created": "2015-10-31T22:22:54.690851953Z",
+        "created_by": "/bin/sh -c #(nop) ADD file:a3bc1e842b69636f9df5256c49c5374fb4eef1e281fe3f282c65fb853ee171c5 in /"
+      },
+      {
+        "created": "2015-10-31T22:22:55.613815829Z",
+        "created_by": "/bin/sh -c #(nop) CMD [\"sh\"]",
+        "empty_layer": true
+      }
+    ]
+}
+`
+)
+
+var (
+	refStr = `{"digest":"<manifest_digest>","mediaType":"application/vnd.oci.image.manifest.v1+json","size":<manifest_size>}`
+
+	manifestStr = `{
+    "annotations": null,
+    "config": {
+        "digest": "<config_digest>",
+        "mediaType": "application/vnd.oci.image.config.v1+json",
+        "size": <config_size>
+    },
+    "layers": [
+        {
+            "digest": "<layer_digest>",
+            "mediaType": "application/vnd.oci.image.layer.tar+gzip",
+            "size": <layer_size>
+        }
+    ],
+    "mediaType": "application/vnd.oci.image.manifest.v1+json",
+    "schemaVersion": 2
+}
+ `
+)
+
+type tarContent struct {
+	header *tar.Header
+	b      []byte
+}
+
+type imageLayout struct {
+	rootDir  string
+	layout   string
+	ref      string
+	manifest string
+	config   string
+	tarList  []tarContent
+}
+
+func TestValidateLayout(t *testing.T) {
+	root, err := ioutil.TempDir("", "oci-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(root)
+
+	il := imageLayout{
+		rootDir:  root,
+		layout:   layoutStr,
+		ref:      refTag,
+		manifest: manifestStr,
+		config:   configStr,
+		tarList: []tarContent{
+			tarContent{&tar.Header{Name: "test", Size: 4, Mode: 0600}, []byte("test")},
+		},
+	}
+
+	// create image layout bundle
+	err = createImageLayoutBundle(il)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = ValidateLayout(root, []string{refTag}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func createImageLayoutBundle(il imageLayout) error {
+	err := os.MkdirAll(filepath.Join(il.rootDir, "blobs", "sha256"), 0700)
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(filepath.Join(il.rootDir, "refs"), 0700)
+	if err != nil {
+		return err
+	}
+
+	// create image layout file
+	err = createLayoutFile(il.rootDir)
+	if err != nil {
+		return err
+	}
+
+	// create image layer blob file.
+	desc, err := createImageLayerFile(il.rootDir, il.tarList)
+	if err != nil {
+		return err
+	}
+	il.manifest = strings.Replace(il.manifest, "<layer_digest>", desc.Digest, 1)
+	il.manifest = strings.Replace(il.manifest, "<layer_size>", strconv.FormatInt(desc.Size, 10), 1)
+
+	desc, err = createConfigFile(il.rootDir, il.config)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


image/manifest_test.go
```diff
@@ -122,3 +103,45 @@ func TestUnpackLayer(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func createTarFile(name string, list []tarContent) (descriptor, error) {
```
wking: We use tar files for image-layout as well.  This helper is about creating layers, so it should probably be `createLayer` or similar to distinguish from image-layout tooling.

wking: And we have a public `Descriptor` type in `github.com/opencontainers/image-spec/specs-go`.  Can we return that instead of the private, local `descriptor` (one step closer to being able to drop the local `descriptor` ;)?

image/manifest_test.go
```diff
@@ -122,3 +103,45 @@ func TestUnpackLayer(t *testing.T) {
...
+		if err = tw.WriteHeader(content.header); err != nil {
+			tw.Close()
+			gw.Close()
+			f.Close()
+			return descriptor{}, err
+		}
+		if _, err = io.Copy(tw, bytes.NewReader(content.b)); err != nil {
+			tw.Close()
+			gw.Close()
+			f.Close()
```
wking: This looks like a job for `defer`  ;).

wking: Perhaps you avoid it because you want to ensure the file is completed and closed before you start hashing it.  You could jump through some hoops to get the `defer`ed functions to fire, but it's probably easier and more efficient to use something like:

```go
file, err := os.Create(name)  // + err checking
hash := sha256.New()
tarMultiWriter := io.MultiWriter(file, hash)
gzipWriter := gzip.NewWriter(tarMultiWriter)
```

image/image_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	err = os.MkdirAll(filepath.Join(root, "blobs", "sha256"), 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = os.MkdirAll(filepath.Join(root, "refs"), 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
```
wking: This image-layout directory initialization also duplicates existing logic in `manifest_test.go`.  Can we pull it out into `createDirectoryImageLayout` or some such?

wking: Actually, it looks like `manifest_test.go` is only stubbing in the `blobs` directory, but not setting up the rest of image-layout.  It would certainly help if we had a `createDirectoryImageLayout` with the more complete logic you have here so we could have a full image-layout there instead of its current partial format.

image/image_test.go
```diff
@@ -0,0 +1,259 @@
...
+	layerPath := filepath.Join(root, "blobs", "sha256", "test.tar")
+
+	desc, err := createTarFile(layerPath, []tarContent{
+		tarContent{&tar.Header{Name: "test", Size: 4, Mode: 0600}, []byte("test")},
+	})
+	if err != nil {
+		return descriptor{}, err
+	}
+
+	err = os.Rename(layerPath, filepath.Join(root, "blobs", "sha256", desc.Digest))
```
wking: `manifest_test.go` is also doing similar post-creation renaming.  Can we use `createLayerFile` there as well?

image/image_test.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = createLayoutFile(root)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = ValidateLayout(root, []string{"latest"}, nil)
```
wking: It's a bit odd to have `latest` explicitly here, but to have it implicit in the `createRefFile` call earlier.  Can we have `createRefFile(root string, ref string, descriptor *specs.Descriptor) err`?

image/manifest_test.go
runcom: > And we have a public Descriptor type in github.com/opencontainers/image-spec/specs-go. Can we return that instead of the private, local descriptor (one step closer to being able to drop the local descriptor ;)?

if it doesn't need to call any of the method attached to `descriptor` - then yes I would use the public `Descriptor` as well - otherwise let's keep `descriptor` for now

wking:
On Fri, Sep 16, 2016 at 01:01:27AM -0700, Antonio Murdaca wrote:
> +func createTarFile(name string, list []tarContent) (descriptor, error) {
> 
> > And we have a public Descriptor type in
> > github.com/opencontainers/image-spec/specs-go. Can we return that
> > instead of the private, local descriptor (one step closer to being
> > able to drop the local descriptor ;)?
> 
> if it doesn't need to call any of the method attached to
> `descriptor` - then yes I would use the public `Descriptor` as well
> - otherwise let's keep `descriptor` for now

I think having public and private descriptor types is more
complication than the convenience of methods is worth.  I've converted
the methods to functions in #5 (among other things ;) if you want to
see how it looks.

xiekeyang:
I improve the patch according to reviewer's comments.
* set up common functions for creating tar file, image layers and hashed digest blobs.
* apply the functions for manifest unit test.
* set up unit test for `ValidateImageLayout`.
* set up configurable reference.
* make these changes in unit test files, not to join binary compiling. If other implementation need them, it is easy to move the functions to functionality files.

PTAL.

cc @wking @runcom @stevvooe @jonboulle

image/manifest_test.go
```diff
@@ -66,58 +54,33 @@ func TestUnpackLayerDuplicateEntries(t *testing.T) {
...
-	tw.WriteHeader(&tar.Header{Name: "test", Size: 4, Mode: 0600})
-	io.Copy(tw, bytes.NewReader([]byte("test")))
-	tw.Close()
-	gw.Close()
-	f.Close()
-
-	// generate sha256 hash
-	h := sha256.New()
-	file, err := os.Open(tarfile)
+	err = os.MkdirAll(filepath.Join(root, "blobs", "sha256"), 0700)
```
wking: You should use `createImageLayoutBundle` here to get `oci-layout`, etc.

image/image_test.go
```diff
@@ -0,0 +1,331 @@
...
+func newDescriptor(name string) (descriptor, error) {
+	file, err := os.Open(name)
+	if err != nil {
+		return descriptor{}, err
+	}
+	defer file.Close()
+
+	// generate sha256 hash
+	hash := sha256.New()
+	size, err := io.Copy(hash, file)
```
wking: This is a second pass through a file that you just streamed to disk.  It would be more efficient to [hash and size while you were streaming to disk][1], but as long as this is just for the test suite a second pass is probably fine.

[1]: https://github.com/opencontainers/image-tools/pull/4#r79103605

image/image_test.go
```diff
@@ -0,0 +1,331 @@
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, err = io.Copy(f, bytes.NewBuffer([]byte(layoutStr)))
+	return err
+}
+
+func createRefFile(root string, mft descriptor) error {
+	refpath := filepath.Join(root, "refs", "latest")
```
wking: `"latest"` → `refTag`.

image/manifest_test.go
xiekeyang: It seems some inappropriate to use `createImageLayoutBundle` here.
the `TestUnpackLayer` prerequisite is just layer file with right path and its descriptor. It even needn't real manifest file. So it is enough to make layer dir and create layer file.
If `TestUnpackLayer` function is missing some test case( I read `testManifest.unpack()` function, and think `TestUnpackLayer` is fine), it is another story.
if it is correct?

image/image_test.go
xiekeyang: fixed.

wking:
On Sat, Sep 17, 2016 at 11:16:24PM -0700, xiekeyang wrote:
> xiekeyang commented on this pull request.
> 
> 
> 
> >  
> -	gw := gzip.NewWriter(f)
> -	tw := tar.NewWriter(gw)
> -
> -	tw.WriteHeader(&tar.Header{Name: "test", Size: 4, Mode: 0600})
> -	io.Copy(tw, bytes.NewReader([]byte("test")))
> -	tw.Close()
> -	gw.Close()
> -	f.Close()
> -
> -	// generate sha256 hash
> -	h := sha256.New()
> -	file, err := os.Open(tarfile)
> +	err = os.MkdirAll(filepath.Join(root, "blobs", "sha256"), 0700)
>
> It seems some inappropriate to use `createImageLayoutBundle` here.
> the `TestUnpackLayer` prerequisite is just layer file with right
> path and its descriptor. It even needn't real manifest file. So it
> is enough to make layer dir and create layer file.

…/blobs/sha256/… is *part* of an image-layout, but not a complete
image layout (e.g. it has no oci-layout file).  If the unpacker
implementation checked for an oci-layout file to confirm “yes, this is
an image-layout 1.0.0.  I can lookup the blob at {algo}/{hash}”, then
it would fail this test as it stands.  I think any unpacker which
expects to be using an image-layout should be checking for that
oci-layout file, and consider it a bug that the unpacker apparently
doesn't check.  If we use createImageLayoutBundle, then this test will
continue to pass after we fix that bug.

xiekeyang:
@wking 

> …/blobs/sha256/… is *part* of an image-layout, but not a complete
image layout (e.g. it has no oci-layout file).  If the unpacker
implementation checked for an oci-layout file to confirm “yes, this is
an image-layout 1.0.0.  I can lookup the blob at {algo}/{hash}”, then
it would fail this test as it stands.  I think any unpacker which
expects to be using an image-layout should be checking for that
oci-layout file, and consider it a bug that the unpacker apparently
doesn't check.  If we use createImageLayoutBundle, then this test will
continue to pass after we fix that bug.

I'm all for you. But I honestly dislike to fix other bugs on this patch.
My preference is to commit only `image_test.go` for image layout validation file on this patch.
And fix manifest unpacking problem and improve the manifest unit test on other patch.
WDYT?

wking:
On Sun, Sep 18, 2016 at 01:09:05AM -0700, xiekeyang wrote:
> But I honestly dislike to fix other bugs on this patch.  My
> preference is to commit only `image_test.go` for image layout
> validation file on this patch.  And fix manifest unpacking problem
> and improve the manifest unit test on other patch.

It's all going to end up in the same stomach ;).  My preference would
be to have multiple commits in this PR, but if you want to punt some
of it to a follow-up PR I'm ok with that.  The (small) risk of
splitting this sort of thing is that the original contributor looses
interest after the PR that partially transitions to the new system,
and other folks have to pick up the pieces (and either revert or
complete the transition) on their behalf.

xiekeyang:
@wking 
I research `func (m *manifest) unpack` related functions, find it is just responsible for unpack layers. Before it is called, finding manifest, finding config and manifest validation are always done. So `manifest.unpack()` needn't to check whole image bundle.
The mechanism may be confused, and need reconstruct, which seems not a bug. I suggest that we had better not to expend this PR scope. And I'd try to improve the mechanism on new PR.
WDYT?

xiekeyang:
cc @runcom @stevvooe @jonboulle @s-urbaniak @wking 
PTAL

wking:
On Sun, Sep 18, 2016 at 02:28:09AM -0700, xiekeyang wrote:
> Before it is called, finding manifest, finding config and manifest
> validation are always done. So `manifest.unpack()` needn't to check
> whole image bundle.

It still needs to pull layers from the image-layout directory, so I
think we want a full image-layout structure.  But we're having trouble
convincing each other, so that seems like a good enough reason to
punt.  As it stands in 966be8fc, I think this PR is generally moving
us in a good direction.  If there's too much in the air to be able to
get to the bottom of our differences here, I'm ok picking up those
discussions in follow-up work.  And if/when #5 lands I think all of
these image/*_test.go will be using the generic engines, so we won't
have to worry about whether a partially-setup image-layout makes sense
or not.

xiekeyang:
cc @stevvooe @runcom @jonboulle

xiekeyang:
As we need to confirm and fix some problem in `func (m *manifest) unpack`, I'd like to narrow this PR, only for function `ValidateLayout`, which I think it is OK. And I'm working on the above discussed issue and will present new PR.

cc @wking @stevvooe

wking:
On Thu, Sep 22, 2016 at 01:19:54AM -0700, xiekeyang wrote:
> As we need to confirm and fix some problem in `func (m *manifest)
> unpack`, I'd like to narrow this PR, only for function
> `ValidateLayout`, which I think it is OK.

f3064f0 → e85687e just dropped all of your changes from
image/manifest_test.go.  I think those were solid, useful changes, so
I don't see a need to drop them here.  But if the image/image_test.go
stuff lands faster because you've dropped them, I'm ok with the
additional punt.

xiekeyang:
@wking Right. hopefully this image test will stand. And I'll submit manifest test.
cc @s-urbaniak @jonboulle @runcom PTAL.

xiekeyang:
PTAL.

vbatts:
LGTM

PR #5, opened 2016-09-16
image: Refactor to use cas/ref engines instead of walkers

wking:
Abstract out the details of the image-layout format, since the validators and unpackers shouldn't care about that sort of backend stuff. This should make it easy to build backends based on zip files, HTTP, the Docker registry format, etc.

Carried from opencontainers/image-spec#159.  The outstanding issue there was what to do about tar handling.  I'm fine with this PR as it stands, but I've [floated a few alternatives][1] in case the maintainers want to pick an alternative path.

[1]: https://github.com/opencontainers/image-spec/pull/159#issuecomment-244303955


wking:
I've pushed 9687484 → 90c4fea with:

* An ‘=’ → ‘:=’ fix in cmd/oci-refs/list.go's printName.
* O_RDWR → O_WRONLY in image/layout/tar.go's CreateTarFile.
* A new commit adding directory-backed engines (to compliment the
  previous tar-backed engines).  The directory-backed engines are much
  nicer ;).

On Fri, Sep 16, 2016 at 01:53:40AM -0700, W. Trevor King wrote:
> Abstract out the details of the image-layout format, since the validators and unpackers shouldn't care about that sort of backend stuff. This should make it easy to build backends based on zip files, HTTP, the Docker registry format, etc.
> 
> Carried from opencontainers/image-spec#159.  The outstanding issue there was what to do about tar handling.  I'm fine with this PR as it stands, but I've [floated a few alternatives][1] in case the maintainers want to pick an alternative path.
> 
> [1]: https://github.com/opencontainers/image-spec/pull/159#issuecomment-244303955
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/opencontainers/image-tools/pull/5
> 
> -- Commit Summary --
> 
>   * Makefile: Add a pattern rule for oci-* commands
>   * image/cas: Add a generic CAS interface
>   * image/refs: Add a generic name-based reference interface
>   * specs-go: Add ImageLayoutVersion and check oci-layout in tar engines
>   * image/layout/tar.go: Add TarEntryByName
>   * image/cas/put: Add a PutJSON helper
>   * image/cas: Implement Engine.Put
>   * vendor: Bundle golang.org/x/net/context
>   * image/*/interface: Add unstable warnings to Engines
>   * image/layout/tar: Add a CreateTarFile helper
>   * image/refs: Implement Engine.Put
>   * image: Refactor to use cas/ref engines instead of walkers
>   * cmd: Document the cas, refs, and init commands
> 
> -- File Changes --
> 
>     M .gitignore (5)
>     M Makefile (20)
>     A cmd/oci-cas/get.go (93)
>     A cmd/oci-cas/main.go (38)
>     A cmd/oci-cas/oci-cas-get.1.md (27)
>     A cmd/oci-cas/oci-cas-put.1.md (27)
>     A cmd/oci-cas/oci-cas.1.md (47)
>     A cmd/oci-cas/put.go (83)
>     M cmd/oci-create-runtime-bundle/main.go (9)
>     M cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md (4)
>     A cmd/oci-image-init/image_layout.go (57)
>     A cmd/oci-image-init/main.go (37)
>     A cmd/oci-image-init/oci-image-init-image-layout.1.md (27)
>     A cmd/oci-image-init/oci-image-init.1.md (33)
>     A cmd/oci-image-tools.7.md (40)
>     M cmd/oci-image-validate/main.go (14)
>     M cmd/oci-image-validate/oci-image-validate.1.md (8)
>     A cmd/oci-refs/get.go (78)
>     A cmd/oci-refs/list.go (75)
>     A cmd/oci-refs/main.go (39)
>     A cmd/oci-refs/oci-refs-get.1.md (27)
>     A cmd/oci-refs/oci-refs-list.1.md (27)
>     A cmd/oci-refs/oci-refs-put.1.md (27)
>     A cmd/oci-refs/oci-refs.1.md (55)
>     A cmd/oci-refs/put.go (81)
>     M cmd/oci-unpack/main.go (13)
>     M cmd/oci-unpack/oci-unpack.1.md (4)
>     M glide.lock (10)
>     M image/autodetect.go (3)
>     A image/cas/interface.go (47)
>     A image/cas/layout/interface.go (25)
>     A image/cas/layout/main.go (37)
>     A image/cas/layout/tar.go (108)
>     A image/cas/put.go (47)
>     M image/config.go (69)
>     M image/descriptor.go (115)
>     M image/image.go (202)
>     A image/layout/layout.go (22)
>     A image/layout/tar.go (267)
>     M image/manifest.go (119)
>     M image/manifest_test.go (45)
>     A image/refs/interface.go (82)
>     A image/refs/layout/main.go (37)
>     A image/refs/layout/tar.go (135)
>     D image/walker.go (118)
>     A vendor/golang.org/x/net/LICENSE (27)
>     A vendor/golang.org/x/net/PATENTS (22)
>     A vendor/golang.org/x/net/context/context.go (156)
>     A vendor/golang.org/x/net/context/go17.go (72)
>     A vendor/golang.org/x/net/context/pre_go17.go (300)
> 
> -- Patch Links --
> 
> https://github.com/opencontainers/image-tools/pull/5.patch
> https://github.com/opencontainers/image-tools/pull/5.diff
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/opencontainers/image-tools/pull/5

-- 
This email may be signed or encrypted with GnuPG (http://www.gnupg.org).
For more information, see http://en.wikipedia.org/wiki/Pretty_Good_Privacy

wking:
Pushed 90c4fea → 7c9efa7 adding:

* oci-cas delete …
* oci-refs delete …
* DirDelete to avoid [1]:

    $ make lint
    checking lint
    image/cas/layout/dir.go:108::warning: duplicate of image/refs/layout/dir.go:156-167 (dupl)
    image/refs/layout/dir.go:156::warning: duplicate of image/cas/layout/dir.go:108-119 (dupl)

[1]: https://travis-ci.org/opencontainers/image-tools/jobs/160621600#L228

xiekeyang:
@wking

This patch seems to be too great to be commited sooner. Could you please split it and submit one by one commit? Now we need the method instead of walking, and go forward to some other functionality based on it. I'm working on my local side [1] referring to your implementation, which is temporary solution and should wait for your patch merged. So I hope your commit can be merged. Thanks so much.

https://github.com/xiekeyang/image-tools/commit/ff0401330fdca591f67b82924c74fc84ce9d052f

wking:
On Wed, Sep 21, 2016 at 05:27:46PM -0700, xiekeyang wrote:
> Could you please split it and submit one by one commit?

First commit submitted separately as #28.  If/when that lands, I'll
file a PR for the next commit, etc. (unless this PR has already landed
in its entirety ;).

wking:
On Wed, Sep 21, 2016 at 10:54:19PM -0700, W. Trevor King wrote:
> On Wed, Sep 21, 2016 at 05:27:46PM -0700, xiekeyang wrote:
> > Could you please split it and submit one by one commit?
> 
> First commit submitted separately as #28.  If/when that lands, I'll
> file a PR for the next commit, etc. (unless this PR has already
> landed in its entirety ;).

Rebased onto master with 7c9efa7 → 1926150 now that #28 has landed.
Bottom of that stack submitted separately as #40 in case that
continues to help review.

wking:
I've pushed 1926150 → ee44a09 which:

* Checks the result of the `DirEngine.Put` cleanup `Remove`, and returns a wrapped error if both the `Put` *and* the cleanup `Remove` fail.
* Adds a comment describing the public `DirDelete` function.

wking:
I've pushed ee44a09 → 4dfe06f fixing a few more lint issues with the
tip commit:

* Cleanup Remove error wrapping in the cas DirEngine.Put (previously
  I'd just fixed the refs DirEngine.Put).
* Check the io.Copy error status in cas DirEngine.Put.
* Don't mask err in cas TarEngine.Put now that it's using blobPath.

cmd/oci-refs/list.go
```diff
@@ -0,0 +1,78 @@
...
+func (state *listCmd) run() (err error) {
+	ctx := context.Background()
+
+	engine, err := layout.NewEngine(ctx, state.path)
+	if err != nil {
+		return err
+	}
+	defer engine.Close()
+
+	return engine.List(ctx, "", -1, 0, state.printName)
```
cyphar: For me this fails with

```
../../.local/src/image-tools/cmd/oci-refs/list.go:69: cannot use state.printName (type func("image-tools/vendor/golang.org/x/net/context".Context, string) error) as type refs.ListNameCallback in argument to engine.List
```

Isn't Go an amazing language? They didn't have vendoring so people rolled their own, then they implemented an incompatible version that breaks if the wind blows in the wrong direction. Absolutely amazing.

cmd/oci-refs/put.go
```diff
@@ -0,0 +1,81 @@
...
+	defer engine.Close()
+
+	decoder := json.NewDecoder(os.Stdin)
+	var descriptor specs.Descriptor
+	err = decoder.Decode(&descriptor)
+	if err != nil {
+		return err
+	}
+
+	return engine.Put(ctx, state.name, &descriptor)
```
cyphar: This fails with 

```
../../.local/src/image-tools/cmd/oci-refs/put.go:80: cannot use &descriptor (type *"image-tools/vendor/github.com/opencontainers/image-spec/specs-go".Descriptor) as type *"github.com/opencontainers/image-tools/vendor/github.com/opencontainers/image-spec/specs-go".Descriptor in argument to engine.Put
```

cmd/oci-refs/list.go
cyphar: In which case, I know why it's happening. It's because I have my gopath setup so that `GOPATH=~` and then in `~/src/` I have symlinks like `github.com/opencontainers -> ~/src` which basically keeps `~/src` less insane. Unfortunately this is not the first time I've had problem. **sigh**

wking:
On Mon, Oct 24, 2016 at 05:58:26AM -0700, Aleksa Sarai wrote:
> The only request I have is that we add some documentation about how all of these tools will fit together to allow for the creation of an image (it's a bit complicated since you need to combine ~5 different tools in order for it to work).

We don't have sufficient tools to create an image until something like #8 lands.  But I agree that having overview docs is good for providing an entry-point for new users.  I have something like that in this PR with [oci-image-tools(7)][1] and there are similar README changes in flight with #48.  If you want something besides those or you want them extended, can you be more specific about what you're looking for?

[1]: https://github.com/opencontainers/image-tools/commit/c8c9708d3a1fc916d64c61e41b05feaf0317a3d4#diff-fc4edae5c4391cf1575e6630b1ba33b5R1

cyphar:
@wking While #48 is good atm, I'd really like to also have just a shell session which shows that you can create an image with **just** this tooling (including #8 once it's merged):

```
% oci-create-layer base root | oci-cas put image
% rm -r base; mv root base; cp -r base root # this can be replaced if we implement #8 using go-mtree.
% # some more create-layer commands
% # some oci-refs commands to show adding refs
% # tar up the image to give you an image tar
```

Then provide some examples of validating and finally unpacking the image we've created (along with the runtime `config.json`. The only thing I can currently see that might be missing from the above is a way to set runtime metadata for image layers. TBH, I'm not **completely** familiar with the `image-spec` but AFAICS that required tooling is still missing. And by having an example script which creates a full image, it's much more clear what tooling needs to be added to the toolking -- as well as being a very useful guide for people who want to use the tooling.

wking:
On Mon, Oct 24, 2016 at 12:46:14PM -0700, Aleksa Sarai wrote:
> While #48 is good atm, I'd really like to also have just a shell session which shows that you can create an image with **just** this tooling (including #8 once it's merged):

I don't talk about the full image-creation logic (because you can't do that without something like #8), but there are currently a few shell-sessions in the docs (e.g. [here][1] and [here][2]).  Do you have shell lines you'd like to see that aren't covered in those examples?

[1]: https://github.com/opencontainers/image-tools/pull/5/commits/c8c9708d3a1fc916d64c61e41b05feaf0317a3d4#diff-eefaad542197c5e2b07a1e5118575c99R34
[2]: https://github.com/opencontainers/image-tools/pull/5/commits/c8c9708d3a1fc916d64c61e41b05feaf0317a3d4#diff-7a76bf9249149df5f157de4af142426dR38

cyphar:
@wking We can add the examples after this PR and #8 is merged. I can't give you a verbatim shell session because I can't compile this PR at the moment :wink: but I would like to have a single shell session inside the man page for `oci-image-tool` which shows the process for going from nothing to a full OCI image using our tooling. Do you understand what I mean? The examples you linked only use single tools, we need to show how the tooling works together.

wking:
On Mon, Oct 24, 2016 at 01:42:11PM -0700, Aleksa Sarai wrote:
> We can add the examples after this PR and #8 is merged… I would like to have a single shell session inside the man page for `oci-image-tool` which shows the process for going from nothing to a full OCI image using our tooling.

That makes sense, and I'm happy to come back and do this after this PR and #8 (or something like them) have landed.

> … we need to show how the tooling works together.

Combinatorics make this expensive to do in general, but going from nothing to an image and then to a runtime bundle is a a limited enough target (once we can actually complete that process).

cmd/oci-refs/list.go
cyphar: I've figured out how to fix it and I'll open a separate PR.

cyphar: I figured it out and opened a PR. #52 fixes the issue.

image/cas/layout/dir.go
```diff
@@ -0,0 +1,122 @@
...
+	engine := &DirEngine{
+		path: path,
+	}
+
+	err = layout.CheckDirVersion(ctx, engine.path)
+	if err != nil {
+		return nil, err
+	}
+
+	engine.temp = filepath.Join(path, "tmp")
```
cyphar: This isn't deleted when I run `oci-cas`. Should there be a defer statement somewhere (not here obviously but somewhere else)? Maybe add a cleanup method for `DirEngine` and defer it in the `oci-cas` caller?

cyphar: Also (bit of a pedantic thing) IMO this should be a randomly generated name because you don't want to intentionally make collisions bad (for example two people modifying an image at the same time). Or if an image setup failed and we didn't clean up, we should avoid messing with broken state.

wking: There is a cleanup method (`Delete`), but closing one CAS/ref engine rooted in this directory doesn't mean we have closed all such engines and can remove this directory.  If we give the directory a random name we could remove it in `Delete`.  Do you mind having multiple random `tmp-******` siblings with random `blob-******` content?

cyphar: Yeah, I think that `tmp-*****` siblings is a better way than just `tmp`. Cheers.

cyphar:
@wking I've been playing with this for a bit and have noticed that `oci-refs` is a bit ... ugly to deal with. Namely, it gives you the reference JSON not the blob that is being referenced. What do you think of adding a `--deref` or `--follow` flag which will take the descriptor in `refs/` and retrieve the blob? The current implementation is really just a wrapper around `rm`, `cat` and `ls` -- IMO it should have "smarter" features.

cyphar:
Also, `oci-refs` really should be able to just take a `digest` and then figure out what the digest is. I was trying to generate images and this part was quite cumbersome -- in order to create a ref I need to by hand create these two things:

1. A modified manifest by hand which I feed into `oci-cas put`. This will be handled by #59.
2. A modified descriptor by hand, which I feed into `oci-ref put`. This should be handled by `oci-ref`. since every object has a `mediaType` it shouldn't be that hard to figure out what the target media type is. I just realised that there's a bug in the spec: a descriptor has a `mediaType` that isn't itself (which makes sense but it means that a descriptor referring to a descriptor will not be able to automatically figure out what the target type is from `mediaType`).

Do you think we should implement 2 before this PR lands or after? As above, I'd be happy to help out with a PR.

wking:
On Wed, Oct 26, 2016 at 06:53:00AM -0700, Aleksa Sarai wrote:
> Also, `oci-refs` really should be able to just take a `digest` and then figure out what the digest is.

I disagree, and would much rather avoid baking type-detection heuristics into the tooling.  See related discussion in opencontainers/image-spec#411, especially [these][1] [comments][2].

> 2. A modified descriptor by hand, which I feed into `oci-ref put`. This should be handled by `oci-ref`. since every object has a `mediaType` it shouldn't be that hard to figure out what the target media type is. I just realised that there's a bug in the spec: a descriptor has a `mediaType` that isn't itself (which makes sense but it means that a descriptor referring to a descriptor will not be able to automatically figure out what the target type is from `mediaType`).

And the config object doesn't have a media type either.  And layer objects aren't even JSON.  There may be UIs we can build that make descriptor initialization easier, but I don't think we want media-type guessing based on blob inspection to be one of them.

[1]: https://github.com/opencontainers/image-spec/pull/411#discussion_r84496898
[2]: https://github.com/opencontainers/image-spec/pull/411#discussion_r84524976

cyphar:
> I disagree, and would much rather avoid baking type-detection heuristics into the tooling.

I don't really think that "look at the mediaType" counts as heuristics. But if you're planning on removing them my honest opinion is that it's a bad idea because it means that you cannot possibly know what a blob represents unless you have the entire image (everything becomes contextual).

> And the config object doesn't have a media type either. And layer objects aren't even JSON. There may be UIs we can build that make descriptor initialization easier, but I don't think we want media-type guessing based on blob inspection to be one of them.

When will there be a reference to a non-descriptor? Is that even valid in the spec?

wking:
On Wed, Oct 26, 2016 at 10:10:37AM -0700, Aleksa Sarai wrote:
> > I disagree, and would much rather avoid baking type-detection
> > heuristics into the tooling.
>
> I don't really think that "look at the mediaType" counts as
> heuristics.

It assumes that the blob you're looking at is JSON that can unmarshal
into Versioned (or @vbatt's new MediaTyped,
opencontainers/image-spec#411).  I'd rather not make that assumption.

> But if you're planning on removing them my honest opinion is that
> it's a bad idea because it means that you cannot possibly know what
> a blob represents unless you have the entire image (everything
> becomes contextual).

You don't need an entire image, you just need a descriptor referencing
the blob in question.  We SHOULD descriptors for all blob references
[1], so I'm comfortable relying on that.  Do you have a workflow where
you're likely to come across a digest but *not* have a full
descriptor?

> > And the config object doesn't have a media type either. And layer
> > objects aren't even JSON. There may be UIs we can build that make
> > descriptor initialization easier, but I don't think we want
> > media-type guessing based on blob inspection to be one of them.
>
> When will there be a reference to a non-descriptor? Is that even
> valid in the spec?

All reference payloads are descriptors [2,3], and those descriptors
can point at whatever you want [4].

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/descriptor.md#L10
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/image-layout.md#L15
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/image-layout.md#L55
[4]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/image-layout.md#L56

cyphar:
On 10/27/2016 04:23 AM, W. Trevor King wrote:
> On Wed, Oct 26, 2016 at 10:10:37AM -0700, Aleksa Sarai wrote:
>> > I disagree, and would much rather avoid baking type-detection
>> > heuristics into the tooling.
>>
>> I don't really think that "look at the mediaType" counts as
>> heuristics.
>
> It assumes that the blob you're looking at is JSON that can unmarshal
> into Versioned (or @vbatt's new MediaTyped,
> opencontainers/image-spec#411). I'd rather not make that assumption.

But tar+gzip has a magic header, which is unique. So there's no issue of 
"I don't know what this binary blob is". Not to mention that all JSON 
blobs we have start with '{'.

>> But if you're planning on removing them my honest opinion is that
>> it's a bad idea because it means that you cannot possibly know what
>> a blob represents unless you have the entire image (everything
>> becomes contextual).
>
> You don't need an entire image, you just need a descriptor referencing
> the blob in question. We SHOULD descriptors for all blob references
> [1], so I'm comfortable relying on that. Do you have a workflow where
> you're likely to come across a digest but *not* have a full
> descriptor?

"The entire image" == "A section of the image's DAG that is greater than 
just the blob". If you can't effectively figure out the type of a JSON 
blob you get from `oci-cas` then *the `oci-cas` tooling is useless*.

Maybe I'm misunderstanding the purpose of this tooling, but from my 
perspective this should be tooling designed to make the spec *not get in 
the way of people trying to create images*. I'm going to end up writing 
a wrapper for all of this, and I'm currently wondering whether I should 
just reimplement all of this instead of reusing it (because it looks 
like the goals of this project don't really match what I would be 
looking for in a useful image toolkit).

>> > And the config object doesn't have a media type either. And layer
>> > objects aren't even JSON. There may be UIs we can build that make
>> > descriptor initialization easier, but I don't think we want
>> > media-type guessing based on blob inspection to be one of them.
>>
>> When will there be a reference to a non-descriptor? Is that even
>> valid in the spec?
>
> All reference payloads are descriptors [2,3], and those descriptors
> can point at whatever you want [4].

My reading of that line is that *currently* it points to an 
image-manifest. And it *may* *in future* change to something else. I 
don't read that line as "a valid image can make a reference point to 
literally anything" -- because I don't see how that would be useful from 
an image PoV (not to mention that `oci-create-runtime-bundle` would 
probably break in those cases).

I'm trying to make the point that in a valid image that this tooling can 
deal with, references will point to image manifests. Those image 
manifests will point to layers and configs. Now, the config isn't a 
self-describing blob but that's not a problem we can solve right now 
(well, we actually could -- by adding a new field that `skopeo` can 
remove -- but then the reference hashes would change between OCI and 
Docker). But that's all separate issues.

I don't see why we should be going backwards -- self-describing blobs 
make things so much easier for consumers.

-- 
Aleksa Sarai
Software Engineer (Containers)
SUSE Linux GmbH
https://www.cyphar.com/

wking:
> If you can't effectively figure out the type of a JSON blob you get from `oci-cas` then *the `oci-cas` tooling is useless*.

`oci-cas` abstracts away the details of the CAS engine.  It could be a tar-backed image-layout 1.0.0, or a directory-backed image-layout 1.0.0, or anything you like as long as you can implement `cas.Engine` over it.  That seems useful to me in its own right.  Folks who want peek-inside media-type detection (or anything else) can build on top of the CAS tooling.

> I don't read that line as "a valid image can make a reference point to literally anything" -- because I don't see how that would be useful from an image PoV (not to mention that `oci-create-runtime-bundle` would probably break in those cases).

I see image-layout as a generic mutable refs + CAS format that happens to be useful for storing images which, along with many other things, fit into the refs + CAS model well.  Keeping refs truly generic makes image extension easy, since users can reference docs, or source code, or their custom manifest format, or whatever.  I agree that generic OCI tooling will not be able to validate/unpack these extensions, but that's fine.  --strick validation could error out if any such extensions are found.

But I think CAS and ref engine abstractions are a useful base on top of which these domain-specific tools can be built.

wking:
When this PR was opencontainers/image-spec#159, @stevvooe pushed back
against the O(N) tar-entry lookup [1], but pointed out that it's hard
to avoid if we handle compressed image-layout tarballs (which we don't
at the moment, but might want to support in the future).  I said I
thought it was fine [2], but the issue is still up in the air [3].
Recently, @cyphar floated some similar tar handling in
containers/image#148 which used an offest cache in tarCache.  @mtrmac
liked the idea, but feels that it's too brittle to use with the
current stdlib's archive/tar.Reader implementation [4].  Do any
maintainers want to weigh in on whether the current O(N) tar-entry
lookups are ok here, or if @cyphar's offset approach is sufficient, or
if this is blocked on some other way around the O(N) problem?

[1]: https://github.com/opencontainers/image-spec/pull/159#r69995233
[2]: https://github.com/opencontainers/image-spec/pull/159#r70000478
[3]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-10-31.log.html#t2016-10-31T22:04:13
[4]: https://github.com/containers/image/pull/148#discussion_r86426030

wking:
On Thu, Nov 03, 2016 at 01:35:59PM -0700, W. Trevor King wrote:
> … some other way around the O(N) problem?

Note that this could be “unpack the tarball to a directory and use the
usual directory-based engine on that”.  This has a higher disk cost
(unless you're unpacking the original tarball on the fly as you read
it off the network) but makes all the other tar-support-sucks issues
go away ;).  I'm willing to do that with the “tar-backed engine” (and
rebuild a new tarball on .Close()?), but again, I'm fine with the
current O(N) tar support and it would be nice to have maintainers
weigh in on the desired approach.

cyphar:
For those who are interested in `imagectl` (which I've currently just called `umoci` as a working name), I've written up a [quick design doc](https://github.com/cyphar/umoci/blob/master/DESIGN.md) on what I expect the final UX to look like (and how it will be implemented). Note that I haven't actually started work on the code yet for multiple reasons (university exams, as well as the fact that `go-mtree` and this project aren't ready yet).

stevvooe:
@cyphar Sorry for the slow cycle, but it looks like that link is broken.

Shall we take this design into an issue and close this PR?

wking:
On Fri, Nov 18, 2016 at 12:11:25PM -0800, Stephen Day wrote:
> @cyphar Sorry for the slow cycle, but it looks like that link is broken.

He removed it in cyphar/umoci@8d19f36.  The last version before that
removal is in [1].

> Shall we take this design into an issue and close this PR?

I'm happy to discuss any design differences between what I have here
and the choices @cyphar has made in umoci.  I don't think ditching all
of this is the best way to have that discussion ;).  Last night saw
some progress on #40, maybe you want to chime in there on any issues
you see with the CAS engine interface or the tar-get implementation?

[1]: https://github.com/cyphar/umoci/blob/ccfb56453111d13452a5198c7831151bdcd0a78b/DESIGN.md

cyphar:
@stevvooe I've removed the file because all of the stuff in the design document has since been implemented in [`umoci`](https://github.com/cyphar/umoci) and the design document no longer accurately described how everything is implemented.

q384566678:
SGTM, What is the current state of this patch?

wking:
> SGTM, What is the current state of this patch?

Wait, is there really renewed interest in this over a year since I filed it as opencontainers/image-spec#159?  If so, you should probably start with #40, which has just the CAS part of this PR.  And getting a green light or concrete improvement suggestions from @stevvooe would be nice, since he's been pushing back against my proposed CAS interface, most recently [here][1].

[1]:  https://github.com/opencontainers/image-tools/pull/40#issuecomment-280499578

stevvooe:
@q384566678 We worked hard on defining a useful content api in https://godoc.org/github.com/containerd/containerd/content. It handles all the cases of resumption, keeps memory usage low and handles pre and post known digest entries, as well as concurrency issues. We've also been able to use `context` to wrap it in transaction managers.

There may be a few tweaks we still need to make, but it is the culmination of work around the old docker/distribution, docker and containerd, as well as some research on distributed content addressable storage systems.

I'd advocate for closing this PR.

Let me know if you have questions.

wking:
> We worked hard on defining a useful content api in https://godoc.org/github.com/containerd/containerd/content.

Comparing that API's [`Provider.Reader`][1]:

```go
type Provider interface {
  Reader(ctx context.Context, dgst digest.Digest) (io.ReadCloser, error)
  ReaderAt(ctx context.Context, dgst digest.Digest) (io.ReaderAt, error)
}
```

with my `Engine.Get`:

```go
type Engine interface {
  Get(ctx context.Context, digest digest.Digest) (reader io.ReadCloser, err error)
}
```

You can see tht we have *exactly the same API*.  And it is also [shared by umoci, where it is `GetBlob`][3].  So I don't see any problem with the part of this PR's API that I've covered in #40.



[1]: https://godoc.org/github.com/containerd/containerd/content#ReadBlob
[2]: https://godoc.org/github.com/containerd/containerd/content#Provider
[3]: https://github.com/openSUSE/umoci/blob/v0.3.0/oci/cas/cas.go#L68

q384566678:
@wking @stevvooe @xiekeyang @cyphar 
Sorry I forgot to reply. So the main question right now is the API question?
 I've seen similar functionality in `umoci`. And these features are important for image-tools propulsion, and I wonder if we can discuss and solve this?

wking:
In case folks are interested, I have more recent work on this front in the stand-alone [casengine][1] repo, and there's a good ref-engine reader in [oci-discovery][2] (although there is currently no ref-engine writer there).  @xiekeyang and I would be [happy to get feedback on those and, eventually if feedback is positive, transfer those to the OCI][3].

[1]: https://github.com/wking/casengine
[2]: https://github.com/xiekeyang/oci-discovery
[3]: https://groups.google.com/a/opencontainers.org/forum/m/#!topic/dev/ZwX6CwrCiK4

PR #9, opened 2016-09-18
Add --version to cmd to show version information

coolljt0725:
Users should have a way to know the spec version and code commit of
the image-tool
````
[lei@fedora image-tools]$ ./oci-create-runtime-bundle --version
commit: 721ba581955c2b8ce1184d7492ed5e3f5ba3e1d4
spec: 0.3.0-dev
[lei@fedora image-tools]$ ./oci-image-validate --version
commit: 721ba581955c2b8ce1184d7492ed5e3f5ba3e1d4
spec: 0.3.0-dev
[lei@fedora image-tools]$ ./oci-unpack --version
commit: 721ba581955c2b8ce1184d7492ed5e3f5ba3e1d4
spec: 0.3.0-dev
````

### DIFF
--- a/cmd/oci-create-runtime-bundle/main.go
+++ b/cmd/oci-create-runtime-bundle/main.go
@@ -20,9 +20,14 @@ import (
 	"os"
 	"strings"
 
+	specs "github.com/opencontainers/image-spec/specs-go"
 	"github.com/opencontainers/image-tools/image"
 	"github.com/spf13/cobra"
 )
+
+// gitCommit will be the hash that the binary was built from
+// and will be populated by the Makefile
+var gitCommit = ""
 
 // supported bundle types
 var bundleTypes = []string{
@@ -31,11 +36,12 @@ var bundleTypes = []string{
 }
 
 type bundleCmd struct {
-	stdout *log.Logger
-	stderr *log.Logger
-	typ    string // the type to bundle, can be empty string
-	ref    string
-	root   string
+	stdout  *log.Logger
+	stderr  *log.Logger
+	typ     string // the type to bundle, can be empty string
+	ref     string
+	root    string
+	version bool
 }
 
 func main() {
@@ -81,10 +87,20 @@ func newBundleCmd(stdout, stderr *log.Logger) *cobra.Command {
 It is strongly recommended to keep the default value.`,
 	)
 
+	cmd.Flags().BoolVar(
+		&v.version, "version", false,
+		`Print version information and exit`,
+	)
 	return cmd
 }
 
 func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
+	if v.version {
+		v.stdout.Printf("commit: %s", gitCommit)
+		v.stdout.Printf("spec: %s", specs.Version)
+		os.Exit(0)
+
+	}
 	if len(args) != 2 {
 		v.stderr.Print("both src and dest must be provided")
 		if err := cmd.Usage(); err != nil {
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 GO15VENDOREXPERIMENT=1
 export GO15VENDOREXPERIMENT
 
+COMMIT=$(shell git rev-parse HEAD 2> /dev/null || true)
+
 EPOCH_TEST_COMMIT ?= v0.2.0
 
 default: help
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by runcom** [comment]
yeah, git commit will be useful also to print some debug while validating images maybe

File: cmd/oci-create-runtime-bundle/main.go [hunk 1]
```diff
@@ -81,10 +87,20 @@ func newBundleCmd(stdout, stderr *log.Logger) *cobra.Command {
...
+		&v.version, "version", false,
+		`Print version information and exit`,
+	)
 	return cmd
 }
 
 func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
+	if v.version {
+		v.stderr.Printf("commit: %s", gitCommit)
+		v.stderr.Printf("spec: %s", specs.Version)
```
wking: I expect image-tools to support a range of spec versions, not just a single one.  For example, once image-spec cuts 2.0, image-tools should still be able to process 1.x images for some reasonable deprecation period.  image-spec is setting us up for that possibility by using the `v1` package for the Go types that are likely to change.  And SemVer means that if we can unpack a 1.3 image, we can also unpack 1.2, 1.1, and 1.0 images.  So I think we want this to be printing:

> specs: 0.4.0

now.  Once opencontainers/image-spec#289 lands and we bump to the new tag, we should be showing:

> specs: 0.5.0

Once 1.0 is cut we should show:

> specs: 1.0.0

Once 1.1 is cut we should show our position in the 1.x series:

> specs: 1.1.0

And then when 2.0 is cut we should show our positions in both the 1.x and 2.x series.  If 2.0 happened after 1.1 (without a 1.2, etc.), that would be:

> specs: 1.1.0, 2.0.0

And if the caller explicitly asked for the version information with `--version`, I think we should be printing it to stdout instead of stderr.

coolljt0725: @wking 
> I expect image-tools to support a range of spec versions, not just a single one. For example, once image-spec cuts 2.0, image-tools should still be able to process 1.x images for some reasonable deprecation period. image-spec is setting us up for that possibility by using the v1 package for the Go types that are likely to change. And SemVer means that if we can unpack a 1.3 image, we can also unpack 1.2, 1.1, and 1.0 images. So I think we want this to be printing:

I think this is hard. It's not only the `v1` package, the spec define the image layout, how we validate the image, how we unpack the layer... etc, this could be differ from version to version. the tool is a implementation of the spec at some point, the 
````
specs: 0.4.0
````
means the tool is developed based on the v0.4.0 spec. 
After 1.0, I agree to show the supported versions but the tool need to do some backward compatibility thing.

>And if the caller explicitly asked for the version information with --version, I think we should be printing it to stdout instead of stderr.

This is a mistake, will fixed.

wking:
On Sun, Sep 18, 2016 at 11:27:36PM -0700, Lei Jitang wrote:
> > I expect image-tools to support a range of spec versions, not just
> > a single one. For example, once image-spec cuts 2.0, image-tools
> > should still be able to process 1.x images for some reasonable
> > deprecation period. image-spec is setting us up for that
> > possibility by using the v1 package for the Go types that are
> > likely to change. And SemVer means that if we can unpack a 1.3
> > image, we can also unpack 1.2, 1.1, and 1.0 images. So I think we
> > want this to be printing:
>
> I think this is hard. It's not only the `v1` package, the spec
> define the image layout, how we validate the image, how we unpack
> the layer... etc, this could be differ from version to version.

And where there are backwards-incompatible changes, media types should
be bumped to represent that.  So the tooling might need to different
unpacking helpers (e.g. one for 1.2 and another for 2.0), but the 1.2
unpacker would be able to handle 1.0, 1.1, and 1.2.  I think we should
include minors in either our media types
(e.g. application/vnd.oci.image.manifest.v1.0+json) or in a version
field inside the JSON (where runtime-spec uses ociVersion [1]) so the
tooling that only handles 1.x through 1.2 can detect and error out if
it gets a 1.3 manifest (or whatever).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#specification-version

vbatts:
i like this as a first step. I do not expect the tool to keep a matrix of compatibility support, at least not in the initial stages. And the addition of project URL can be added later.

LGTM

PR #12, opened 2016-09-20
reduplicated condition

xiekeyang:
I read all `walkFunc` in `(*pathWalker).walk(f walkFunc)`, which are all deal with directory FileInfo condition. Here seems to be a reduplicated block.

### DIFF
--- a/image/walker.go
+++ b/image/walker.go
@@ -74,12 +74,6 @@ loop:
 	return nil
 }
 
-type eofReader struct{}
-
-func (eofReader) Read(_ []byte) (int, error) {
-	return 0, io.EOF
-}
-
 type pathWalker struct {
 	root string
 }
@@ -101,10 +95,6 @@ func (w *pathWalker) walk(f walkFunc) error {
 		rel, err := filepath.Rel(w.root, path)
 		if err != nil {
 			return errors.Wrap(err, "error walking path") // err from filepath.Walk includes path name
-		}
-
-		if info.IsDir() { // behave like a tar reader for directories
-			return f(rel, info, eofReader{})
 		}
 
 		file, err := os.Open(path)


glestaris:
@xiekeyang I don't understand why/if this works. 

Is the walker going to find directories in the process of reading the path? 

If so, how comes `Read` works on the file object that `os.Open` returns when the path is a directory? To me, it looks like `Read` returns an error when the file object is a directory.

xiekeyang:
@glestaris  return result is finally decided by `walkFunc`, which return `nil` at the begin when path is directory. It can be done after `os.Open()`. 
cc @stevvooe @wking @runcom

stevvooe:
@s-urbaniak PTAL

s-urbaniak:
I am :-1: on this change. While it is true that callers currently always check in `FileInfo` whether the given file object is a directory, this changes makes the returned reader in `pathWalker` inconsistent with the behavior of `tarWalker`.

`tarWalker` delegates to [tar#Reader](https://godoc.org/archive/tar#Reader) which in case of a directory returns a valid `io.Reader`, see https://golang.org/src/archive/tar/reader.go#L818, always returning 0 bytes read and an immediate `io.EOF`.

`pathWalker` in this PR would return a reader err'ing when trying to `io#Reader.Read` from a directory, making both walker behaviors inconsistent.

xiekeyang:
@s-urbaniak 
Got it and thanks for your explain. So it is for possible `reader.Read()` case inside `walkFunc`?
However, at the beginning of each `walkFunc`, it always check:
```go
        if info.IsDir() || (...) {
                return nil
        }
```
It seems tarWalker and pathWalker have same behavior of `return nil`. 
If I misunderstand, or the case `f(rel, info, eofReader{})` doesn't work?

s-urbaniak:
@xiekeyang not sure I am following. The `walkFunc` could `reader.Read()` in case of `info.IsDir() == true`. The only thing this code does is to enforce consistent behavior. You are right that we are jumping out anyways, but I remember in my initial implementation I had one case where the `info.IsDir()` check was not part of every `walkFunc`.

My suggestion to keep it there, it doesn't really hurt, and only maintains consistency between `tarWalker`, and `pathWalker`.

xiekeyang:
@s-urbaniak Understood. I'll close the PR.

PR #26, opened 2016-09-21
Display correct runtime-spec version

coolljt0725:
Instead of using a constant as the runtime-spec
version, user `Version` in the package.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -70,7 +70,7 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 	}
 
 	var s specs.Spec
-	s.Version = "0.5.0"
+	s.Version = specs.Version
 	// we should at least apply the default spec, otherwise this is totally useless
 	s.Process.Terminal = true
 	s.Root.Path = rootfs


runcom:
lgtm (ianam)

image/config.go
```diff
@@ -70,7 +70,7 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 	}
 
 	var s specs.Spec
-	s.Version = "0.5.0"
+	s.Version = specs.Version
```
runcom: this was the runtime version not the image one right? if so this is wrong (do image and runtime follow same release? if so this is still wrong to me, but righter...)

wking:
Dup of #9 (which I think is doing a more thorough job)?

wking:
On Thu, Sep 22, 2016 at 09:43:18AM -0700, Antonio Murdaca wrote:
> > @@ -70,7 +70,7 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
>  	}
>  
>  	var s specs.Spec
> -	s.Version = "0.5.0"
> +	s.Version = specs.Version
> 
> this was the runtime version not the image one right? if so this is
> wrong (do image and runtime follow same release? if so this is still
> wrong to me, but righter...)

I think this change is right.  It's using the runtime version
(specs.Version) to fill in a field on the generated runtime config.
With this PR, there's one less place to bump when image-tools moves to
new runtime-spec versions.  Although with the approach made possible
by opencontainers/runtime-tools#219, we will be able to drop the
runtime-spec dependency completely.

vbatts:
LGTM

PR #28, opened 2016-09-22
Makefile: Add a pattern rule for oci-* commands

wking:
Make it easy to (re)build a single command without building all of them.  Using main.go as the only prerequisite cuts some corners, but prerequisites for this are hard ;).  With this commit, you can touch `main.go` to force a rebuild.  If that ends up being too awkward, we can make these `.PHONY` targets.

This is the current first commit in #5, to see if [nibbling away at that PR one commit at a time][1] makes review any easier.

[1]: https://github.com/opencontainers/image-tools/pull/5#issuecomment-248781012

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,10 @@ export GO15VENDOREXPERIMENT
 COMMIT=$(shell git rev-parse HEAD 2> /dev/null || true)
 
 EPOCH_TEST_COMMIT ?= v0.2.0
+TOOLS := \
+	oci-create-runtime-bundle \
+	oci-image-validate \
+	oci-unpack
 
 default: help
 
@@ -20,10 +24,10 @@ check-license:
 	@echo "checking license headers"
 	@./.tool/check-license
 
-tools:
-	go build -ldflags "-X main.gitCommit=${COMMIT}" ./cmd/oci-create-runtime-bundle
-	go build -ldflags "-X main.gitCommit=${COMMIT}" ./cmd/oci-unpack
-	go build -ldflags "-X main.gitCommit=${COMMIT}" ./cmd/oci-image-validate
+tools: $(TOOLS)
+
+oci-%: cmd/oci-%/main.go
+	go build -ldflags "-X main.gitCommit=${COMMIT}" ./cmd/$@
 
 lint:
 	@echo "checking lint"
@@ -69,10 +73,7 @@ install.tools: .install.gitvalidation .install.glide .install.glide-vc .install.
 	gometalinter --install --update
 
 clean:
-	rm -rf *~ $(OUTPUT_DIRNAME)
-	rm -f oci-create-runtime-bundle
-	rm -f oci-unpack
-	rm -f oci-image-validate
+	rm -rf *~ $(OUTPUT_DIRNAME) $(TOOLS)
 
 .PHONY: \
 	tools \
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ check-license:
 
 tools: $(TOOLS)
 
-oci-%: cmd/oci-%/main.go
+$(TOOLS): oci-%:
 	go build -ldflags "-X main.gitCommit=${COMMIT}" ./cmd/$@
 
 lint:
@@ -77,6 +77,7 @@ clean:
 
 .PHONY: \
 	tools \
+	$(TOOLS) \
 	check-license \
 	clean \
 	lint \


runcom:
> If that ends up being too awkward, we can make these .PHONY targets.

yes please, split them into individual targes and then have "tools" to build them all

wking:
On Thu, Sep 22, 2016 at 09:40:40AM -0700, Antonio Murdaca wrote:
> > If that ends up being too awkward, we can make these .PHONY targets.
>
> yes please, split them into individual targes and then have "tools"
> to build them all

As is stood in 0179710, this PR already split them into individual
(pattern) targets and had the ‘tools’ target build them all.  What it
did not do was complete dependency checking (which is tricky with Go)
to only rebuild when necessary.  I've pushed 0179710 → 689ee3e with a
new commit that adds .PHONY to always rebuild.  That rebuilds too
often, but it errs on the side of rebuilding while 0179710 errs on the
side of not-rebuilding.

Makefile
```diff
@@ -67,13 +71,11 @@ install.tools: .install.gitvalidation .install.glide .install.glide-vc .install.
...
 clean:
-	rm -rf *~ $(OUTPUT_DIRNAME)
-	rm -f oci-create-runtime-bundle
-	rm -f oci-unpack
-	rm -f oci-image-validate
+	rm -rf *~ $(OUTPUT_DIRNAME) $(TOOLS)
 
 .PHONY: \
 	tools \
+	$(TOOLS) \
```
vbatts: that is somewhat redundant

vbatts: also, not true. When those binaries are built from the subdirectory, then the executable by those names is present.

Makefile
```diff
@@ -18,10 +22,10 @@ check-license:
 	@echo "checking license headers"
 	@./.tool/check-license
 
-tools:
-	go build ./cmd/oci-create-runtime-bundle
-	go build ./cmd/oci-unpack
-	go build ./cmd/oci-image-validate
+tools: $(TOOLS)
+
+$(TOOLS): oci-%:
```
vbatts: Here the oci-% is looking for dependencies of the target that have the oci-% pattern. Is that intended?

wking:
On Thu, Sep 22, 2016 at 10:50:54AM -0700, Vincent Batts wrote:
> >  
>  .PHONY: \
>  	tools \
> +	$(TOOLS) \
> 
> that is somewhat redundant

We need $(TOOLS) so we rebuild them every time.  You could remove
‘tools’ if you wanted, but there is no ‘tools’ file on the disk, so I
think keeping it in .PHONY is appropriate.

wking:
On Thu, Sep 22, 2016 at 10:56:01AM -0700, Vincent Batts wrote:
> +$(TOOLS): oci-%:
>
> Here the oci-% is looking for dependencies of the target that have
> the oci-% pattern. Is that intended?

This is a static pattern rule (linked from the the 689ee3e92 commit
message).  It is saying “for all the targets in $(TOOLS) use the
‘oci-%: ’ pattern rule”.  It has no dependencies (entries after the
second colon).

vbatts:
would almost seem like to get the Makefile dependency usefulness, we could have something like
```Makefile
GO_DEPS = $(shell go list -f '{{range .Deps}}{{printf "%s" .}}{{end}}' $(1))
GO_SOURCE_FILES = $(shell go list -f '{{range .GoFiles}}{{printf "%s" .}}{{end}}' $(1))

oci-%: $(call GO_SOURCE_FILES,$(call GO_DEPS,./cmd/oci-%))
        go build ./cmd/$@
```
or similar

wking:
On Thu, Sep 22, 2016 at 11:27:41AM -0700, Vincent Batts wrote:
> would almost seem like to get the Makefile dependency usefulness, we could have something like
> ```Makefile
> GO_DEPS = $(shell go list -f '{{range .Deps}}{{printf "%s" .}}{{end}}' $(1))
> GO_SOURCE_FILES = $(shell go list -f '{{range .GoFiles}}{{printf "%s" .}}{{end}}' $(1))
> 
> oci-%: $(call GO_SOURCE_FILES,$(call GO_DEPS,./cmd/oci-%))

Having an accurate list of dependencies would be awesome, but the
.GoFiles line seems to print the filename and not the full path to the
file.  I'm happy to add this sort of checking if you know the proper
incantation to get the full path.

On the other hand, ‘go build’ (as we call it without -a) only rebuilds
packages that are out of date, so I'm fine leaving this PR at
689ee3e924, calling ‘go build’ every time, and having it bail out
if/when dependent packages are up to date.

vbatts:
yes. `go build` is nicer.

wking:
On Thu, Sep 22, 2016 at 11:48:06AM -0700, Vincent Batts wrote:
> yes. `go build` is nicer.

So we're both happy with this PR as it stands with 689ee3e9?  Or is
there anything else I can do to help move this along?

vbatts:
please rebase

wking:
Rebased around #9 with 689ee3e → 2d06368.

vbatts:
LGTM

vbatts:
welp https://travis-ci.org/opencontainers/image-tools/builds/165582442

wking:
On Thu, Oct 06, 2016 at 09:35:21AM -0700, Vincent Batts wrote:
> welp https://travis-ci.org/opencontainers/image-tools/builds/165582442

Is that lint timeout reproducible?  I'd just kick it again.

vbatts:
funky. that did the trick.

PR #29, opened 2016-09-22
descriptor validation

xiekeyang:
Function `(d *descriptor) validate()` is for descriptor variable, with which the exact cas file path is clear. So in this function, it is safely to open the target file, needn't to walk again.
Other walking mechanism is working properly, to need overall refactor, so I don't touch them hastily.

This PR refers to #5 , by @wking, on reading file part. I'm not very sure if this PR is valuable to be accepted individually. Otherwise we should close it and wait for #5 , or please @wking amend this PR with his singed-off?

As to me, I hope a cas reading/writing API can be used to code. 

Signed-off-by: xiekeyang <keyang.xie@gmail.com>

cc @wking @stevvooe

### DIFF
--- a/image/reader.go
+++ b/image/reader.go
@@ -1,0 +1,84 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import (
+	"archive/tar"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+type reader interface {
+	Get(desc descriptor) (io.ReadCloser, error)
+}
+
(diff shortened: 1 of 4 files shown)


image/reader.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (r *tarReader) Get(desc descriptor) (io.ReadCloser, error) {
+	f, err := os.Open(r.name)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	tr := tar.NewReader(f)
+loop:
```
vbatts: is the named loop entirely needed? I'm not opposed to them, but I don't see any inner loops to distinguish from

xiekeyang: @vbatts 
```go
loop:
        for {
                hdr, err := tr.Next()
                switch err {
                case io.EOF:
                        break loop
                case nil:
                // success, continue below
                default:
                        return nil, err 
                }   
                if hdr.Name == filepath.Join("blobs", desc.algo(), desc.hash()) &&
                        !hdr.FileInfo().IsDir() {
                        buf, err := ioutil.ReadAll(tr)
                        if err != nil {
                                return nil, err 
                        }   
                        return ioutil.NopCloser(bytes.NewReader(buf)), nil 
                }   
        }   
```

`break` is located in switch-case block, but it want to break the for-loop. So it should point the label( as `break loop`). Otherwise, the break will jump out switch-case block.
Of course, we can use if checking instead of `switch-case`, which not use label. But it is same effect.
If my explain has covered your question?

xiekeyang:
PTAL.

image/reader.go
vbatts: You're right. My apologies.

vbatts:
LGTM

PR #31, opened 2016-09-30
Add make install support

WeiZhang555:
Add `install` support to make, for installing image tools binaries into
$PATH.

Signed-off-by: Zhang Wei zhangwei555@huawei.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_install:
   - go get -u github.com/alecthomas/gometalinter
   - gometalinter --install --update
   - go get -t -d ./...
+  - mkdir -p /tmp/image-tools
 
 install: true
 
@@ -22,3 +23,4 @@ script:
   - make check-license
   - make test
   - make tools
+  - make install DESTDIR=/tmp/image-tools
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TOOLS := \
 	oci-create-runtime-bundle \
 	oci-image-validate \
 	oci-unpack
+DESTDIR ?= /usr/bin/
 
 default: help
 
@@ -19,6 +20,7 @@ help:
 	@echo " * 'lint' - Execute the source code linter"
 	@echo " * 'test' - Execute the unit tests"
 	@echo " * 'update-deps' - Update vendored dependencies"
+	@echo " * 'install' - Install the binaries into system path"
 
 check-license:
 	@echo "checking license headers"
@@ -43,6 +45,9 @@ update-deps:
 	glide-vc --only-code --no-tests
 	# see http://sed.sourceforge.net/sed1line.txt
 	find vendor -type f -exec sed -i -e :a -e '/^\n*$$/{$$d;N;ba' -e '}' "{}" \;
+
+install:
+	install -D -m 755 -t ${DESTDIR} ${TOOLS}
 
 .PHONY: .gitvalidation
 
@@ -81,4 +86,5 @@ clean:
 	check-license \
 	clean \
 	lint \
-	test
+	test \
+	install


Makefile
```diff
@@ -38,6 +40,11 @@ update-deps:
 	# see http://sed.sourceforge.net/sed1line.txt
 	find vendor -type f -exec sed -i -e :a -e '/^\n*$$/{$$d;N;ba' -e '}' "{}" \;
 
+install:
+	install -D -m 755 oci-create-runtime-bundle ${INSTALL_DIR}
+	install -D -m 755 oci-unpack ${INSTALL_DIR}
+	install -D -m 755 oci-image-validate ${INSTALL_DIR}
```
wking: I'm not excited about repeatedly listing all of our compiled tools.  I don't see a way around that here, but with #28 we could do:

    install: $(TOOLS)
        install -D -t $(INSTALL_DIR) $^

WeiZhang555: Seems better. I can wait until #28 is merged. :smile:

vbatts: well, more $(DESTDIR) which is more the autoconf standard

.travis.yml
```diff
@@ -22,3 +22,4 @@ script:
   - make check-license
   - make test
   - make tools
+  - sudo make install
```
vbatts: do we need to have sudo here, if we just add a DESTDIR variable so it can be installed to a limited user tmp dir. Maybe. Just thinking out loud here.

wking: .+1 to using `DESTDIR` and dropping sudo.

WeiZhang555:
>do we need to have sudo here, if we just add a DESTDIR variable so it can be installed to a limited user tmp dir. Maybe. Just thinking out loud here.

@vbatts @wking 
Well, the .travis file and Makefile has switched to ${DESTDIR} as you suggested :smile:

.travis.yml
```diff
@@ -7,12 +7,14 @@ sudo: required
...
 before_script:
   - export PATH=$HOME/gopath/bin:$PATH
+  - export DESTDIR=/tmp/image-tools
 
 before_install:
   - make install.tools
   - go get -u github.com/alecthomas/gometalinter
   - gometalinter --install --update
   - go get -t -d ./...
+  - mkdir /tmp/image-tools -p
```
WeiZhang555: Updated~

wking:
b7c33d85 looks good to me.

WeiZhang555:
Any suggestion? @vbatts

.travis.yml
```diff
@@ -7,12 +7,14 @@ sudo: required
 
 before_script:
   - export PATH=$HOME/gopath/bin:$PATH
+  - export DESTDIR=/tmp/image-tools
```
vbatts: This works, but I guess I would likely just include this with the command

.travis.yml
```diff
@@ -22,3 +24,4 @@ script:
   - make check-license
   - make test
   - make tools
+  - make install
```
vbatts: right here, like `make install DESTDIR=/tmp/image-tools`

WeiZhang555: OK for me.

WeiZhang555:
@vbatts Updated.

.travis.yml
```diff
@@ -22,3 +23,4 @@ script:
   - make check-license
   - make test
   - make tools
+  - DESTDIR=/tmp/image-tools make install
```
WeiZhang555: All right, it's quite convincible. 
I'll update, thanks :smile:

.travis.yml
```diff
@@ -22,3 +23,4 @@ script:
   - make check-license
   - make test
   - make tools
+  - make install DESTDIR=/tmp/image-tools
```
wking: Travis reports [trailing whitespace][1].

[1]: https://travis-ci.org/opencontainers/image-tools/jobs/170663842#L227

WeiZhang555: My bad. It's green now :smile:

wking:
b5e5da53 looks good to me :).

WeiZhang555:
ping @vbatts

xiekeyang:
LGTM

xiekeyang:
LGTM

WeiZhang555:
Since this has been dangling for so long, I'm closing this.

wking:
On Tue, Jan 10, 2017 at 01:17:45AM -0800, zhangwei_cs wrote:
> Since this has been dangling for so long, I'm closing this.

The diff still looks good to me.  I imagine that maintainers are just
busy with other things, and will eventually get back around to
reviewing image-tools PRs.

WeiZhang555:
So maybe I should re-open this and keep it for some more time :joy:

xiekeyang:
LGTM

xiekeyang:
Need Rebase

PR #32, opened 2016-09-30
Error out if rootfs type is not layers

coolljt0725:
The image spec requries that
````
MUST be set to layers. Implementations MUST generate an error if
they encounter a unknown value while verifying or unpacking an image.
```
The image-tool should act as the image spec says.

Signed-off-by: Lei Jiang <leijitang@huawei.com>

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -53,6 +53,10 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 		if err := json.Unmarshal(buf, &c); err != nil {
 			return err
 		}
+		// check if the rootfs type is 'layers'
+		if c.RootFS.Type != "layers" {
+			return fmt.Errorf("'%s' is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
+		}
 		return errEOW
 	}); err {
 	case nil:


image/config.go
```diff
@@ -53,6 +53,10 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 		if err := json.Unmarshal(buf, &c); err != nil {
 			return err
 		}
+		// check if the rootfs type is 'layers'
+		if c.RootFS.Type != "layers" {
+			return fmt.Errorf("'%s' is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
```
wking: `'%s'`→ `%q`.

vbatts:
LGTM

q384566678:
it's checked in JSON Schema,need to check it here？

wking:
On Thu, Oct 20, 2016 at 11:57:24PM -0700, Zhou Hao wrote:
> it's checked in JSON Schema…

It is checked in JSON Schema since opencontainers/image-spec#358, which landed a day after this PR.  So once we bump the vendored image-spec dependency to pull in opencontainers/image-spec#358 we can revert this PR.  I think we want a policy for handling these image-spec validation bugs cleanly going forward (opencontainers/image-spec#406).

q384566678:
> It is checked in JSON Schema since opencontainers/image-spec#358, which landed a day after this PR. 
I think these fixed values（Such as SchemaVersion and MediaType,and so on） should be checked in JSON Schema.

q384566678:
I think these fixed values（Such as SchemaVersion and MediaType,and so on） should be checked in JSON Schema.

wking:
On Fri, Oct 21, 2016 at 12:20:36AM -0700, Zhou Hao wrote:
> I think these fixed values (Such as SchemaVersion and MediaType,and so on) should be checked in JSON Schema.

And in image-spec's master, they are.  But in image-tool's master ([pulling in][1] opencontainers/image-spec@7e6e2f76d6, v1.0.0-rc1~18), they are not (and are checked by the code which landed via this PR).  Once image-spec cuts v1.0.0.rc2 and image-tools bumps its dependency to pull in that release, we can drop the local check.

[1]: https://github.com/opencontainers/image-tools/blob/f24d27b5e62acb27a688e36cbac13902bfb75234/glide.yaml#L4

PR #35, opened 2016-10-06
image/config: Replace `'%s'` with `%q`

wking:
As I [suggested earlier][1].  Fewer characters for us, and safer quoting for consumers.

[1]: https://github.com/opencontainers/image-tools/pull/32#discussion_r81357093


jonboulle:
lgtm

PR #37, opened 2016-10-07
description: validates manifest when unpack

xiekeyang:
Actually, the validation object before unpack is whole manifest schema, not only its own media type.


xiekeyang:
@wking 
I update it, to present the description as,

`oci-unpack` validates an application/vnd.oci.image.manifest.v1+json, with entire Merkle tree rooted on it, and unpacks its layered filesystem to `dest`.

If it is OK?

wking:
“with” → “and the”, but otherwise e0a2f66 looks good to me.

xiekeyang:
.@wking I change `with` to `and the`, although I fell `and.., and` not very smooth.

wking:
f256b57f looks good to me.  Maybe the “…, and …, and” would feel better to you if we use parens?

> `oci-unpack` validates an application/vnd.oci.image.manifest.v1+json (and the entire Merkle tree rooted on it) and unpacks its layered filesystem to `dest`.

vbatts:
LGTM

q384566678:
Please rebase.

q384566678:
@xiekeyang Please rebase.

PR #38, opened 2016-10-08
find manifest and config file by cas path

xiekeyang:
To use CAS reader directly to find and get manifest and config file, instead of `walkFunc` to whole directory.
cc @wking @vbatts 
Signed-off-by: xiekeyang xiekeyang@huawei.com

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -18,9 +18,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


image/config.go
```diff
@@ -34,38 +32,34 @@ import (
 type config v1.Image
 
 func findConfig(w walker, d *descriptor) (*config, error) {
-	var c config
-	cpath := filepath.Join("blobs", d.algo(), d.hash())
+	var (
+		c     config
+		cpath = filepath.Join("blobs", d.algo(), d.hash())
+	)
```
wking: I don't see a point in bumping this.

xiekeyang: @wking Do you only mean `cpath` is redundant because it is used once? I fixed it.

xiekeyang: Oh, I'd follow the popular repos custom as what you said, and fixed it.

wking:
e7d902cf9 looks good to me.

image/config.go
```diff
@@ -37,35 +35,29 @@ func findConfig(w walker, d *descriptor) (*config, error) {
...
+	if err != nil {
+		return nil, err
+	}
 
-		if err := schema.MediaTypeImageConfig.Validate(bytes.NewReader(buf)); err != nil {
-			return errors.Wrapf(err, "%s: config validation failed", path)
-		}
+	buf, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
```
coolljt0725: one nit, I think it's better to keep the error message consistent, can we use 
`return nil, errors.Wrapf(err, "%s: error reading config", path)` here?

xiekeyang: Fix in https://github.com/opencontainers/image-tools/pull/38/commits/62165b47a9fa59d16e056c09a068461e1f61fabe#diff-60e1f0f301bd2d474f51924e35e52c64R45

PR #39, opened 2016-10-11
Add config validation

q384566678:
I think the config may be validated when verifying the image file

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -64,19 +64,42 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	}
 }
 
-func (c *config) validate(w walk) error {
+func (c *config) validate() error {
 
 	// check if the rootfs type is 'layers'
 	if c.RootFS.Type != "layers" {
-		return errors.New("%q is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
+		return fmt.Errorf("%q is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
 	}
 
-	for _, d = range c.RootFs.DiffIDs {
-		h := sha256.New()
-		DiffIds := "sha256" + hex.EncodeToString(h.Sum(nil))
+	//check os and architecture
+	if err := c.checkPlatform(); err != nil {
+		return err
+	}
+	return nil
+}
 
-		if DiffIds != d {
-			return errors.New("DiffIDs mismatch")
+func (c *config) checkPlatform() error {
+	validCombins := map[string][]string{
+		"darwin":    {"386", "amd64", "arm", "arm64"},
+		"dragonfly": {"amd64"},
+		"freebsd":   {"386", "amd64", "arm"},
+		"linux":     {"386", "amd64", "arm", "arm64", "ppc64", "ppc64le", "mips64", "mips64le"},
+		"netbsd":    {"386", "amd64", "arm"},
+		"openbsd":   {"386", "amd64", "arm"},
+		"plan9":     {"386", "amd64"},
+		"solaris":   {"amd64"},
+		"windows":   {"386", "amd64"}}
+	for os, archs := range validCombins {
+		if os == c.OS {
+			for _, arch := range archs {
+				if arch == c.Architecture {
+					return nil
+				}
+			}
+			return fmt.Errorf("Combination of %q and %q is invalid.", c.OS, c.Architecture)
+		}
+	}
+	return fmt.Errorf("Operation system %q of the bundle is not supported yet.", c.OS)
 }
 
 func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
--- a/image/config.go
+++ b/image/config.go
@@ -53,10 +53,6 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 		if err := json.Unmarshal(buf, &c); err != nil {
 			return err
 		}
-		// check if the rootfs type is 'layers'
-		if c.RootFS.Type != "layers" {
-			return fmt.Errorf("%q is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
-		}
 		return errEOW
 	}); err {
 	case nil:
@@ -66,6 +62,21 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	default:
 		return nil, err
 	}
+}
+
+func (c *config) validate(w walk) error {
+
+	// check if the rootfs type is 'layers'
+	if c.RootFS.Type != "layers" {
+		return errors.New("%q is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
+	}
+
+	for _, d = range c.RootFs.DiffIDs {
+		h := sha256.New()
+		DiffIds := "sha256" + hex.EncodeToString(h.Sum(nil))
+
+		if DiffIds != d {
+			return errors.New("DiffIDs mismatch")
 }
 
 func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
--- a/image/image.go
+++ b/image/image.go
@@ -87,6 +87,16 @@ func validate(w walker, refs []string, out *log.Logger) error {
 		if err := m.validate(w); err != nil {
 			return err
 		}
+
+		c, err := findConfig(w, &m.Config)
+		if err != nil {
+			return err
+		}
+
+		if err := c.validate(w); err != nil {
+			return err
+		}
+
 		if out != nil {
 			out.Printf("reference %q: OK", ref)
 		}
--- a/image/image.go
+++ b/image/image.go
@@ -84,7 +84,7 @@ func validate(w walker, refs []string, out *log.Logger) error {
 			return err
 		}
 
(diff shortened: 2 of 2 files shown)


image/config.go
```diff
@@ -68,6 +66,25 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	}
 }
 
+func (c *config) validate(w walker) error {
+
+	// check if the rootfs type is 'layers'
+	if c.RootFS.Type != "layers" {
+		return errors.New("RootFs is an unknown rootfs type, MUST be 'layers'")
+	}
```
wking: We don't need to check this in Go, because it's checked in JSON Schema since opencontainers/image-spec#358.

q384566678: ok,thanks for your reminder

wking:
On Thu, Oct 13, 2016 at 01:03:42AM -0700, Zhou Hao wrote:
> Closed #39.

You were doing more here than just shifting the RootFS.Type check [1].
I'm not entirely clear on what the rest of the changes are doing, but
you may want to re-open this PR if you feel any of them still need to
be addressed.

[1]: https://github.com/opencontainers/image-tools/pull/39#discussion_r82927806

runcom:
> I think the config may be validated when verifying the image file

sure, but this is just checking os and arch which isn't adding a proper config validation

q384566678:
Maybe I had a problem with the title name. I think config need to verify the place including rootfs (has been verified)), followed by OS and arch, I put the config validation into the findconfig function, and not a separate list of functions @runcom

wking:
On Thu, Oct 13, 2016 at 02:56:14AM -0700, Antonio Murdaca wrote:
> Also, why checking the config while finding it? should it be
> validated in the caller after finding it? it really seems part of
> "validating", not the lookup for the config file.

+1.  Retrieval and validation should be separate steps (although the
current `findConfig` sets a bad example here).

image/config.go
```diff
@@ -68,6 +73,30 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	}
 }
 
+func checkPlatform(c config) error {
```
wking: The desired location for this logic is still up in the air (as far as I can tell), with some folks arguing for just JSON Schema validation in image-spec ([me][1] and [maybe @vbatts][2]) and some (maybe [@philips][3] and [@stevvooe][4]) arguing for as much JSON validation as you can get (JSON Schema or otherwise) without needing to look outside the given JSON blob (e.g.  at children it references).  With the latter approach, this platform check would be landing in the image-spec repo (somewhere under `schema/`?).  It's probably best to wait until @stevvooe has filed his [proposed package structure][5] before moving ahead with this..

[1]: https://github.com/opencontainers/image-spec/pull/337
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.log.html#l-57
[3]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.log.html#l-71
[4]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.log.html#l-74
[5]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.html

q384566678:
ok,I have separated retrieval and validation,
As for the other uncertain arguments, we wait for the results slowly

wking:
On Thu, Oct 13, 2016 at 07:41:57PM -0700, Zhou Hao wrote:
> ok,I have separated retrieval and validation,

The JSON Schema validation [1] should be moved to your new (c
*config)validate() too.

[1]: https://github.com/opencontainers/image-tools/pull/39/files#diff-60e1f0f301bd2d474f51924e35e52c64R49

q384566678:
@wking Thank you for your suggestion, but I do not think so. The JSON Schema validation is to confirm that the format conforms to config, in order to confirm the found file is config file, after which to verify its specific content. In the mainfest.go file, the JSON Schema validation is also placed in the findManifest function, rather than in the validate function

wking:
On Thu, Oct 13, 2016 at 11:13:58PM -0700, Zhou Hao wrote:
> The JSON Schema validation is to confirm that the format conforms to
> config, in order to confirm the found file is config file, after
> which to verify its specific content.

The JSON Schema check is validation that we want to apply to any
config, regardless of how the caller got their hands on the config.
And I don't see why you'd need to use JSON Schema checks to validate
*every config* extracted via findConfig.  If you're following a
trusted descriptor, you can be fairly confident that the referenced
config is a config without the JSON Schema check.  If you are
concerned that the descriptor might not point at a valid config, you
should run *all* of the validation (JSON Schema and otherwise) on the
blob.  And really, “validates against the spec's
application/vnd.oci.image.config.v1+json” is the least of your worries
if you don't trust the descriptor.  You should be more worried about
spec-valid configs that are setting up malicious containers.

q384566678:
We went on to talk about the problem, according to you, the JSON Schema validation should be moved to validate function, if so, should the findManifest function of JSON Schema validation into the corresponding also validate function? My understanding of this aspect is not enough in-depth, thank you for your guidance @wking

wking:
On Tue, Oct 18, 2016 at 11:43:28PM -0700, Zhou Hao wrote:
> … according to you, the JSON Schema validation should be moved to
> validate function, if so, should the findManifest function of JSON
> Schema validation into the corresponding also validate function?

I think all validation should be moved into validation-specific
functions which folks can call when they see fit.  That applies to all
media types, although I don't think we have to clean up all of that in
this PR ;).

Once #5 or something like it lands, I'd like to switch to
callback-based validation and unpacking, where we have a registry of
validators (or unpackers, or whatever) with entries for each supported
media type.  I've sketched this out in [1] if you want to take a look.

[1]: https://github.com/wking/image-tools/commit/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6

image/config.go
```diff
@@ -68,6 +64,44 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	}
 }
 
+func (c *config)validate() error {
```
stevvooe: Please Go format your code.

@vbatts @philips Do we have a Go style guide in place?

stevvooe:
This seems like it belongs upstream in the spec validation code.

q384566678:
> This seems like it belongs upstream in the spec validation code.

But I think it also needs to be verified here

wking:
On Wed, Oct 19, 2016 at 07:12:47PM -0700, Zhou Hao wrote:
> > This seems like it belongs upstream in the spec validation code.
> 
> But I think it also needs to be verified here

You can still verify it here.  There are several possible approaches
to splitting validation between image-tools and image-spec.  @stevvooe
seems to have convinced @vbatts and formed a consensus around keeping
all intra-blob JSON validation in image-spec [1].  To fit that
approach, we need to port this platform check (and likely all of
image/config.go that deals with validation) to image-spec (as
schema/config.go?).  Once that lands in image-spec, we can bump the
image-spec dependency here and call the image-spec validators where we
need intra-blob validation.

[1]: https://github.com/opencontainers/image-tools/pull/39#discussion_r83317029

Mashimiao:

wking:
[@steevooe][1] and [I][2] are pretty sure `checkPlatform` belongs in image-spec.

[1]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-255001231
[2]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-255001231

q384566678:
@wking I think it is very difficult to  check Platform in the image-spec, because os and arch are a range of values rather than a fixed value.

wking:
> I think it is very difficult to check Platform in the image-spec, because os and arch are a range of values rather than a fixed value.

I don't understand.  You should just be able to copy your function over into image-spec.  I can work up a PR for that if it would help clarify.

stevvooe:
This validation component should be using `image-spec` as an upstream. Is there something that I am not seeing that is preventing this from occurring?

q384566678:
@stevvooe In the image-spec, the verification is to rely on json file to verify, and json file can only verify the specific value and type, os and arch are uncertain values (in the array of values ), So in the json file validation only verify its type, and its value is not within the scope of the specified array is difficult to achieve (at least for me, can not be achieved). This is where I have been puzzled.

wking:
On Sun, Nov 20, 2016 at 07:00:58PM -0800, Zhou Hao wrote:
> In the image-spec, the verification is to rely on json file to
> verify, and json file can only verify the specific value and type…

A lot of image-spec validation is based on JSON Schema, but image-spec
has some other checks as well.  See the landed
opencontainers/image-spec#341 and the in-flight
opencontainers/image-spec#403 for examples.

q384566678:
@wking I got it,I understand the wrong before, thank you

stevvooe:
@q384566678 It sounds to me like json-schema is a big bust. It can't seem to do the most basic of validation. More complex cases, like this, are impossible.

stevvooe:
Closing in favor of opencontainers/image-spec#469

PR #40, opened 2016-10-11
image/cas: Add a generic CAS interface

wking:
And implement that interface for tarballs based on the specs image-layout.

Also add a new `oci-cas` command so folks can access the new read functionality from the command line.

The `Context` interface follows the pattern recommended [here][1], allowing callers to cancel long running actions (e.g. push/pull over the network for engine implementations that communicate with a remote store).

This is the current first commit in #5, to see if [nibbling away at that PR one commit at a time][2] makes review any easier.  I have a directory-based backend in #5 if you want to see what that looks like.  In another commit in #5 I replace the `image/walker.go` functionality with this new API.

[1]: https://blog.golang.org/context
[2]: https://github.com/opencontainers/image-tools/pull/5#issuecomment-248781012

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/oci-cas
 /oci-create-runtime-bundle
-/oci-unpack
 /oci-image-validate
+/oci-unpack
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ COMMIT=$(shell git rev-parse HEAD 2> /dev/null || true)
 
 EPOCH_TEST_COMMIT ?= v0.2.0
 TOOLS := \
+	oci-cas \
 	oci-create-runtime-bundle \
 	oci-image-validate \
 	oci-unpack
--- a/cmd/oci-cas/get.go
+++ b/cmd/oci-cas/get.go
@@ -1,0 +1,93 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/opencontainers/image-tools/image/cas/layout"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+)
+
+type getCmd struct {
+	path   string
+	digest string
+}
+
+func newGetCmd() *cobra.Command {
+	state := &getCmd{}
+
+	return &cobra.Command{
+		Use:   "get PATH DIGEST",
+		Short: "Retrieve a blob from the store",
+		Long:  "Retrieve a blob from the store and write it to stdout.",
+		Run:   state.Run,
+	}
+}
+
+func (state *getCmd) Run(cmd *cobra.Command, args []string) {
+	if len(args) != 2 {
+		fmt.Fprintln(os.Stderr, "both PATH and DIGEST must be provided")
+		if err := cmd.Usage(); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+		}
+		os.Exit(1)
+	}
+
+	state.path = args[0]
+	state.digest = args[1]
+
+	err := state.run()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
+	os.Exit(0)
+}
+
+func (state *getCmd) run() (err error) {
+	ctx := context.Background()
+
+	engine, err := layout.NewEngine(state.path)
+	if err != nil {
+		return err
+	}
+	defer engine.Close()
+
+	reader, err := engine.Get(ctx, state.digest)
+	if err != nil {
+		return err
+	}
+	defer reader.Close()
+
+	bytes, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return err
+	}
+
+	n, err := os.Stdout.Write(bytes)
+	if err != nil {
+		return err
+	}
+	if n < len(bytes) {
+		return fmt.Errorf("wrote %d of %d bytes", n, len(bytes))
+	}
+
+	return nil
+}
--- a/cmd/oci-cas/main.go
+++ b/cmd/oci-cas/main.go
@@ -1,0 +1,37 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	cmd := &cobra.Command{
+		Use:   "oci-cas",
+		Short: "Content-addressable storage manipulation",
+	}
+
+	cmd.AddCommand(newGetCmd())
+
+	err := cmd.Execute()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
--- a/image/cas/interface.go
+++ b/image/cas/interface.go
@@ -1,0 +1,44 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cas implements generic content-addressable storage.
+package cas
+
+import (
+	"io"
+
+	"golang.org/x/net/context"
+)
+
+// Engine represents a content-addressable storage engine.
+type Engine interface {
+
+	// Put adds a new blob to the store.  The action is idempotent; a
+	// nil return means "that content is stored at DIGEST" without
+	// implying "because of your Put()".
+	Put(ctx context.Context, reader io.Reader) (digest string, err error)
+
+	// Get returns a reader for retrieving a blob from the store.
+	// Returns os.ErrNotExist if the digest is not found.
+	Get(ctx context.Context, digest string) (reader io.ReadCloser, err error)
+
+	// Delete removes a blob from the store. The action is idempotent; a
+	// nil return means "that content is not in the store" without
+	// implying "because of your Delete()".
+	Delete(ctx context.Context, digest string) (err error)
+
+	// Close releases resources held by the engine.  Subsequent engine
+	// method calls will fail.
+	Close() (err error)
+}
--- a/image/cas/layout/interface.go
+++ b/image/cas/layout/interface.go
@@ -1,0 +1,25 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package layout
+
+import (
+	"io"
+)
+
+// ReadWriteSeekCloser wraps the Read, Write, Seek, and Close methods.
+type ReadWriteSeekCloser interface {
+	io.ReadWriteSeeker
+	io.Closer
+}
--- a/image/cas/layout/main.go
+++ b/image/cas/layout/main.go
@@ -1,0 +1,36 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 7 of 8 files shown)


hustcat:
IMO, we should add `List` API to return all digest for `Engine interface`, such as:

```
List(ctx context.Context) (digest []string, err error)
```

This is needed for [GC](https://github.com/opencontainers/image-tools/issues/86#issuecomment-261323339) in the future.

hustcat:
And more, I'd like to add `Exist` API:

```
// Exist Return true if digest is exist, else return false
Exist(ctx context.Context, digest string) (bool,  error)
```

wking:
On Thu, Nov 17, 2016 at 07:49:24PM -0800, Ye Yin wrote:
> IMO, we should add `List` API to return all digest for `Engine
> interface`, such as:
> 
> ```
> List(ctx context.Context) (digest []string, err error)
> ```
> 
> This is needed for GC in the future.

Listing is expensive, and I'd rather have GC be an opaque function
[1].  But we can certainly revisit this if/when the rest of the CAS
API lands, when it will be easier to push concrete GC implementations.
In the mean time, leaving list off is safer, because adding List is a
backwards compatible change while removing it is a backwards
incompatible change.

[1]: https://github.com/cyphar/umoci/commit/667271b57db5637a18b0f249026a84608f834b3d#commitcomment-19711529

wking:
On Thu, Nov 17, 2016 at 08:25:25PM -0800, Ye Yin wrote:
> And more, I'd like to add `Exist` API:
> 
> ```
> // Exist Return true if digest is exist, else return false
> Exist(ctx context.Context, digest string) (bool,  error)
> ```

Is this for “I want to put this blob, whose hash I already know, but
don't want to bother if that blob is already in storage”?  Maybe in
the context of copying between CAS engines?  And you think that
calling Get and immediately closing any succesfully returned reader
would be too expensive (e.g. if the engine was backed by HTTP and you
could do a HEAD instead of a GET)?  I think there might be a solid
case in there, but I'd rather wait and get the more obviously critical
portions of the API landed first.  Then we can circle back and look at
this sort of optimization in a second pass.

hustcat:
> Is this for “I want to put this blob, whose hash I already know, but
don't want to bother if that blob is already in storage”?  Maybe in
the context of copying between CAS engines?  And you think that
calling Get and immediately closing any succesfully returned reader
would be too expensive ...

Yeah, exactly:)

hustcat:
> Listing is expensive, and I'd rather have GC be an opaque function
[1].  But we can certainly revisit this if/when the rest of the CAS
API lands, when it will be easier to push concrete GC implementations.
In the mean time, leaving list off is safer, because adding List is a
backwards compatible change while removing it is a backwards
incompatible change...

IMO, `List` is a very basic API, `GC` is one case. Get all digests is very common requirement, such as statistics, just like list all references. Whether expensive is related with underlay implementation. For directory walker, maybe not a expensive thing?

wking:
On Thu, Nov 17, 2016 at 10:46:44PM -0800, Ye Yin wrote:
> IMO, `List` is a very basic API, `GC` is one case. Get all digests
> is very common requirement, such as statistics, just like list all
> references…

Listing all references is a necessary evil ;).  To collect statistics,
you can use the ref-engine's List to walk refs (or subsets thereof),
and then walk the Merkle DAG down from each ref to discover blobs.  If
you want a particular blob to show up in lists (and not be garbage
collected), just add a new ref to pin it.  I don't see a use case for
listing blobs in the absence of refs outside of GC.

> Whether expensive is related with underlay implementation. For
> directory walker, maybe not a expensive thing?

For a small, local, directory-backed CAS engine, blob-list performance
would be acceptable.  However, my goal is to write a generic API so
higher-level tools can manipulate CAS without having to care about the
engine implementation.  If your CAS engine involves millions of blobs
accessed via a distributed hash table or sharded across many hosts,
walking significant chunks of that is going to be expensive.  And
garbage collection is likely to be more sophisticated than
stop-the-world mark and sweep.  Unless we have a very convincing use
case for *all* CAS engines to provide a List API, I'd rather avoid it.

And again, the CAS API I'm proposing here covers all the use-cases we
currently have inside image-tools.  If we land this without List and
later decide that List is necessary, a follow-up PR can add it later.
It should only matter for this PR if your argument is that you need
List for a use-case this repository already addresses.

xiekeyang:
@wking I'm reading it but not debug yet. 

Can we use `Get(ctx context.Context, desc descriptor)`, to instead Get(ctx context.Context, digest string). I select `descriptor` as parameter in `image/reader.go` because it bring helpful information, like:
`if hdr.Name == filepath.Join("blobs", desc.algo(), desc.hash())` instead your:
```go
targetName := fmt.Sprintf("./blobs/%s/%s", algorithm, hash)
if header.Name == targetName{
}
```

wking:
On Fri, Nov 18, 2016 at 12:20:55AM -0800, xiekeyang wrote:
> Can we use `Get(ctx context.Context, desc descriptor)`, to instead
> Get(ctx context.Context, digest string).

I've avoided going with the image-spec Descriptor because while size
information may be useful (e.g. you may want to compare the expected
size against the retrieved blob), the media type information is not
useful (the CAS engine should not care or need to track the types of
stored blobs).  I'd *really* like to avoid the image-tools descriptor,
because:

* It's currently private, so I *can't* use it from image/cas.
* Converting between image-spec and image-tools descriptor types is
  an unnecessary headache for users.
* The desc.algo() and desc.hash() helpers are only really interesting
  to CAS-engine implementers.  For everyone else, the digest might as
  well be an opaque string.

In a later #5 commit I add blobPath [1] and refPath [2] helpers to
abstract these out.  If it would make you more comfortable with this
commit I can move those definitions over here.

[1]: https://github.com/opencontainers/image-tools/pull/5/commits/512176d6de3d6082f474c0673c077fd084d73563#diff-c3a5563e81df7aa2a94d70642ba7c3ceR46
[2]: https://github.com/opencontainers/image-tools/pull/5/commits/512176d6de3d6082f474c0673c077fd084d73563#diff-65bd80b0d4865c33446593b6c8e353f3R46

xiekeyang:
> In a later #5 commit I add blobPath [1] and refPath [2] helpers to
abstract these out.  If it would make you more comfortable with this
commit I can move those definitions over here.

Good!

xiekeyang:
@wking 
> Also add a new oci-cas command so folks can access the new read functionality from the command line.

I think CAS engine interface is great, but I'm not sure that command `oci-cas get` is useful. Do we really need this command? (I might misunderstand ` the new read functionality`)

wking:
On Fri, Nov 18, 2016 at 03:34:16AM -0800, xiekeyang wrote:
> > In a later #5 commit I add blobPath [1] and refPath [2] helpers to
> > abstract these out.  If it would make you more comfortable with
> > this commit I can move those definitions over here.
> 
> Good!

Is that “yes, squash those into this commit”?  Or is it “ah, then this
commit is ok because after we chew through the rest of #5 we'll be in
a good place”?

xiekeyang:
> Is that “yes, squash those into this commit”?  Or is it “ah, then this
commit is ok because after we chew through the rest of #5 we'll be in
a good place”?

Sorry, I mean could you please squash those of `blobPath` and `refPath` to this commit?

wking:
On Fri, Nov 18, 2016 at 03:40:35AM -0800, xiekeyang wrote:
> I think CAS engine interface is great, but I'm not sure that command
> `oci-cas get` is useful.

As #5 continues, oci-cas fills in to cover the rest of the interface
(put and delete).  Again, I'm happy to shuffle things around if you'd
like all of that pulled into this commit, or all of the command line
stuff pushed out into a later commit.

> Do we really need this command?

“Really need” is hard to put a finger on.  None of our Go code shells
out to oci-cas ;).  But the command line wrappers are easy to write,
and I like providing a non-Go API for folks who are interested in
interacting with OCI-compatible CAS engines.  I certainly don't think
oci-cas is going to be the highest-level command line tool we offer,
but I don't see a reason to *not* offer it.  For an analogy in another
plumbing-rich context, see Git's cat-file [1] and hash-object [2].

[1]: https://git-scm.com/docs/git-cat-file
[2]: https://git-scm.com/docs/git-hash-object

wking:
On Fri, Nov 18, 2016 at 03:57:19AM -0800, xiekeyang wrote:
> Sorry, I mean could you please squash those of `blobPath` and
> `refPath` to this commit?

blobPath shifted in with 23d8fb7 → fbae6d9.  The ref interface is in a
follow-up commit, but I've adjusted all of #5 to take this
early-*Path-helper approach if you want to check that out.

xiekeyang:
In general I agree this PR. cc @stevvooe WDYT?

stevvooe:
@xiekeyang I don't think this is the right model. It doesn't have a transaction interface for verifying and setting blobs. Half of the interface is useless for the application of tar files and I am not sure what problems this solves for image-tools.

For this to be useful, we'd have to refactor the entire thing. As I said when this PR was in image, please use the docker/distribution repo, and the myriad other Go CAS projects, to inform the API.

wking:
On Wed, Nov 30, 2016 at 03:21:10PM -0800, Stephen Day wrote:
> It doesn't have a transaction interface for verifying and setting
> blobs.

What sort of interface are you looking for?  Setting blobs seems like
it would be covered by Engine.Put [1], and you don't need a
transactional *interface* for a CAS engine (although you do want a
transactional *implementation*).

> Half of the interface is useless for the application of tar files…

Which half?  I can split Engine into CASReader and CASWriter or some
such if that would make the tar backing more appealing.

> … and I am not sure what problems this solves for image-tools.

It abstracts away the details of the CAS engine so we don't need to
bake it in (like we currently do, including walker-based lookup for
directory-backed CAS [2,3]!).

> For this to be useful, we'd have to refactor the entire thing.

I'm happy to refactor.  I've looked at the docker/distribution and
camlistore interfaces [4], and this is where I ended up.  But if you'd
like the interface methods modified, let me know what you'd like
changed and why.

[1]: https://github.com/opencontainers/image-tools/pull/40/files#diff-a372a96ac413817b9f8f567e310b60bcR27
[2]: https://github.com/opencontainers/image-tools/blob/421458f7e467ac86175408693a07da6d29817bf7/image/walker.go#L90-L97
[3]: https://github.com/opencontainers/image-tools/blob/421458f7e467ac86175408693a07da6d29817bf7/image/image.go#L28-L32
[4]: https://github.com/opencontainers/image-spec/pull/159#issuecomment-231517752

PointMicro:
What's the progress on this? I could really use this right about now.

stevvooe:
@PointMicro This approach has some _major_ design problems. What are you trying to do exactly? Maybe I can point your towards something useful.

wking:
On Wed, Feb 15, 2017 at 07:42:08PM -0800, Stephen Day wrote:
> This approach has some _major_ design problems.

I'm still not clear on what these design problems are, but I'm happy
to fix them if you point them out.

Now that we have go-digest as a dependency (#104, #107), I've rerolled
this PR to use the Digest type for digests with fbae6d9 → a44de0d.

stevvooe:
> I'm still not clear on what these design problems are, but I'm happy
to fix them if you point them out.

And I am not going to sit down and reiterate them again. I have provided several examples, there are multiple projects to draw from (umoci, containerd, docker/distribution, camlistore, etc.) that all have addressed these issues.

Furthermore, it treats tar archives as a random access storage medium and makes multiple poor assumptions.

wking:
On Thu, Feb 16, 2017 at 02:57:16PM -0800, Stephen Day wrote:
> … there are multiple projects to draw from (umoci…

umoci has addressed tar access by erroring out when pointed at a
tar-backed image-layout [1].  It's CAS Get interface [2] is identical
to mine [3].

> … containerd,…

I see no image-layout support in containerd:

  containerd $ git grep blobs v0.2.5
  v0.2.5:vendor/src/golang.org/x/sys/unix/syscall_linux.go:               // to be uninterpreted fixed-size binary blobs--but
  v0.2.5:vendor/src/golang.org/x/sys/unix/syscall_solaris.go:             // to be uninterpreted fixed-size binary blobs -- but

Very recently (with docker/containerd#452) you gave containerd a
fetcher which adds a hints argument and makes the ID more generic [4],
but is otherwise identical to my interface.

> … docker/distribution, …

distribution has the same interface, except that it returns
ReadSeekClosers instead of my ReadClosers [5].  It also supports a
getter that returns a byte array [6], but that's clearly the
convenient special case because it would be very expensive for large
blobs.

I couldn't find any image-layout support in docker/distribution (which
isn't surprising, since it's basically an alternative store
format/service).

> … camlistore,…

camlistore seems to dispense with the Context, and it also returns a
size (in addition to the ReadCloser and error) [7].

> Furthermore, it treats tar archives as a random access storage
> medium and makes multiple poor assumptions.

So as far as I can tell, I'm doing the same thing as everyone else,
unless you're really missing camlistore's returned size or
containerd's hints.

Or are you fine with my API, and just concerned with the
implementation?  I don't see anyone else implementing tar-backed
image-layout support, but please point me at them if I've missed them.

[1]: https://github.com/openSUSE/umoci/blob/v0.1.0/oci/cas/cas.go#L133-L137
[2]: https://github.com/openSUSE/umoci/blob/v0.1.0/oci/cas/cas.go#L89-L91
[3]: https://github.com/opencontainers/image-tools/pull/40/files#diff-a372a96ac413817b9f8f567e310b60bcR33
[4]: https://github.com/docker/containerd/pull/452/files#diff-aba5bb73e69236c8e862c4a985731536R9
[5]: https://github.com/docker/distribution/blob/v2.6.0/blobs.go#L147-L150
[6]: https://github.com/docker/distribution/blob/v2.6.0/blobs.go#L144-L145
[7]: https://github.com/camlistore/camlistore/blob/dc66fe281ecd848a12bc102ffdf491033b0454fd/pkg/client/get.go#L47

stevvooe:
@wking The problem here is that you are too focused on "image layout" which is a minor part of the image format.

> Or are you fine with my API, and just concerned with the
implementation?  

No. You're API is naive, racy and memory intensive.

> I don't see anyone else implementing tar-backed
image-layout support, but please point me at them if I've missed them.

Exactly. This makes no sense. The only way to do this with a CAS store would be to have a transactional addition of blobs. Effectively, you have a transaction for each blob addition, then a transaction that creates the tar file.

wking:
On Thu, Feb 16, 2017 at 03:34:29PM -0800, Stephen Day wrote:
> > Or are you fine with my API, and just concerned with the
> > implementation?  
> 
> No. You're API is naive, racy and memory intensive.

The Get API?  It's basically the same as everyone else's [1].

> The only way to do this with a CAS store would be to have a
> transactional addition of blobs.

There is no blob addition going on in this PR.  I've just pushed
ccab818 → 38fe4fc which drops the Put and Delete stubs in case they
were what you were pushing back against.

[1]: https://github.com/opencontainers/image-tools/pull/40#issuecomment-280498988

wking:
I've just pushed 5df5883 → fe6a35b rebasing this work onto master, since @q384566678 [expressed renewed interest][1] in #5, of which this PR is the next component.  I didn't make any changes to `image/cas`, and most of the rebase was adjusting to the `oci-image-tools` consolidation from #103.

[1]: https://github.com/opencontainers/image-tools/pull/5#issuecomment-318273924

wking:
And I've just pushed fe6a35b → cda19fe fixing some `go fmt` issues and replacing the [deprecated `os.SEEK_SET`][1] with [`io.SeekStart`][2] to [fix this lint error][3].

[1]: https://golang.org/pkg/os/#pkg-constants
[2]: https://golang.org/pkg/io/#pkg-constants
[3]: https://travis-ci.org/opencontainers/image-tools/jobs/258217375#L235

wking:
I've pushed cda19fe → 0c9e84e fixing a [missing error check][1].

[1]: https://travis-ci.org/opencontainers/image-tools/jobs/258219874#L227

PR #42, opened 2016-10-18
image/manifest: Recursively remove pre-existing entries when unpacking

### LINKED ISSUE #41: oci-image-tool: error extracting layer: symlink /lib/systemd/systemd tmp2/bin/systemd: file exists
Hello, I am trying to examine a docker images without a docker tool. I have created a simple smallest docker image, where I encounter following error. The image can be found in docker repository: `dohnto/debian-skopeo`. The dockerfile is following

``` Dockerfile
FROM debian:8.3
RUN apt-get update -y && \
    apt-get upgrade -y 
```

**Note**, that during building such an image, a `systemd` will be upgraded:

```
...
The following packages will be upgraded:
  apt base-files dmsetup dpkg e2fslibs e2fsprogs gnupg gpgv libapt-pkg4.12
  libc-bin libc6 libcomerr2 libdevmapper1.02.1 libgcrypt20 libpam-modules
  libpam-modules-bin libpam0g libpcre3 libss2 libsystemd0 libudev1
  multiarch-support perl-base systemd systemd-sysv tzdata udev
27 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
...
```

My workflow in further examination is following:

``` bash
$ skopeo --version
skopeo version 0.1.17-dev commit: 507d09876d0620d46de7c8c20eb9b58d2dffa482
$ skopeo copy docker://dohnto/debian-skopeo oci:tmp
$ oci-image-tool validate tmp
reference "latest": OK
tmp: OK
$ oci-image-tool unpack --ref latest tmp tmp2
unpacking failed: error extracting layer: symlink /lib/systemd/systemd tmp2/bin/systemd: file exists
$ ls -l tmp2/bin/systemd
lrwxrwxrwx 1 tom tom 20 15. říj 16.38 tmp2/bin/systemd -> /lib/systemd/systemd
```

I expected this to work and the `oci-image-tool` to actually unpack the oci image into a folder I could chroot.  Thank you for any help or workaround how to solve this issue.

Version information:
- OS: fedora23 workstation, kernel 4.7.3-100.fc23.x86_64
- running under systemd
- skopeo: 0.1.17-dev commit: 507d09876d0620d46de7c8c20eb9b58d2dffa482
- oci-image-tool: build from source: v0.5.0-63-g32a3cb9

```
```

wking:
Implementing the logic that is in-flight with opencontainers/image-spec#317, but [using recursive removal](https://github.com/opencontainers/image-spec/pull/317/files#r79214718).  GNU tar has a `--recursive-unlink` option that's not enabled by default, with the motivation being something like “[folks would be mad if we blew away a full tree and replaced it with a broken symlink](https://www.gnu.org/software/tar/manual/html_node/Dealing-with-Old-Files.html)”.  That makes sense for working filesystems, but we're building the rootfs from scratch here so losing information is not a concern.  This commit always uses recursive removal to get that old thing off the filesystem (whatever it takes ;).

The exception to the removal is if both the tar entry and existing path occupant are directories.  In this case we want to use GNU tar's default `--overwrite-dir` behavior, but `unpackLayer`'s metadata handling is currently very weak (the in-flight #3 strengthens it a bit) so I've left it at “don't delete the old directory”.

The reworked directory case also fixes a minor bug from 44210d05 (opencontainers/image-spec#177) where the:

```
if fi, err := os.Lstat(path); !(err == nil && fi.IsDir()) {
```

block would not error out if the `Lstat` failed for a reason besides the acceptable `IsNotExist`.  Instead, it would attempt to call `MkdirAll`, which would probably fail for the same reason that `Lstat` failed (ENOTDIR?).  But it's better to handle the `Lstat` errors directly.

Fixes #41.

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -195,11 +195,27 @@ loop:
 			continue loop
 		}
 
+		if hdr.Typeflag != tar.TypeDir {
+			err = os.RemoveAll(path)
+			if err != nil && !os.IsNotExist(err) {
+				return err
+			}
+		}
+
 		switch hdr.Typeflag {
 		case tar.TypeDir:
-			if fi, err := os.Lstat(path); !(err == nil && fi.IsDir()) {
-				if err2 := os.MkdirAll(path, info.Mode()); err2 != nil {
-					return errors.Wrap(err2, "error creating directory")
+			fi, err := os.Lstat(path)
+			if err != nil && !os.IsNotExist(err) {
+				return err
+			}
+			if os.IsNotExist(err) || !fi.IsDir() {
+				err = os.RemoveAll(path)
+				if err != nil && !os.IsNotExist(err) {
+					return err
+				}
+				err = os.MkdirAll(path, info.Mode())
+				if err != nil {
+					return err
 				}
 			}
 
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -160,101 +160,15 @@ loop:
 			return errors.Wrapf(err, "error advancing tar stream")
 		}
 
-		hdr.Name = filepath.Clean(hdr.Name)
-		if !strings.HasSuffix(hdr.Name, string(os.PathSeparator)) {
-			// Not the root directory, ensure that the parent directory exists
-			parent := filepath.Dir(hdr.Name)
-			parentPath := filepath.Join(dest, parent)
-			if _, err2 := os.Lstat(parentPath); err2 != nil && os.IsNotExist(err2) {
-				if err3 := os.MkdirAll(parentPath, 0755); err3 != nil {
-					return err3
-				}
-			}
-		}
-		path := filepath.Join(dest, hdr.Name)
-		if entries[path] {
-			return fmt.Errorf("duplicate entry for %s", path)
-		}
-		entries[path] = true
-		rel, err := filepath.Rel(dest, path)
+		var whiteout bool
+		whiteout, err = unpackLayerEntry(dest, hdr, tr, &entries)
 		if err != nil {
 			return err
 		}
-		info := hdr.FileInfo()
-		if strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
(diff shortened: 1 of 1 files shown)


wking:
I've pushed an additional bb40ed1f to address the cyclomatic complexity warning, and Travis is happy now.

image/manifest.go
```diff
@@ -195,11 +195,27 @@ loop:
...
+			}
+			if os.IsNotExist(err) || !fi.IsDir() {
+				err = os.RemoveAll(path)
+				if err != nil && !os.IsNotExist(err) {
+					return err
+				}
+				err = os.MkdirAll(path, info.Mode())
+				if err != nil {
+					return err
 				}
```
xiekeyang: A debatable view: if it is easier to understand to use below?
```go
fi, err := os.Lstat(path)
if err==nil{
   ...
}else if  os.IsNotExist(err){
  ...
}else if err!=nil
  ...
}
```
And, if `fi, err := os.Lstat(path)` can be replaced to upper lines?
```go
fi, err := os.Lstat(path)
if err == nil && !fi.IsDir() {
  remove path
}else if !os.IsNotExist(err){
  return err 
}   
```

wking: I think both of your proposals will end up duplicating the `MkdirAll` call.  And idiomatic Go seems to be “check for and exit on unrecoverable errors as quickly as possible” (which is how I handle it now with the `if err != nil && !os.IsNotExist(err)`).  But I don't really care.  If the maintainers feel that avoiding `!`s in the conditionals is worth pushing the unrecoverable error to the end of the `case` block, I can reroll to do so.

q384566678:
LGTM

q384566678:
cc @coolljt0725 @xiekeyang

q384566678:
ping @coolljt0725 @xiekeyang

coolljt0725:
LGTM ping @opencontainers/image-tools-maintainers

PR #45, opened 2016-10-20
Increase the validation of the Manifest content

q384566678:
I think it is also necessary to validate these two fixed values in the Manifest

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -33,8 +33,10 @@ import (
 )
 
 type manifest struct {
-	Config descriptor   `json:"config"`
-	Layers []descriptor `json:"layers"`
+	SchemaVersion int          `json:"schemaVersion"`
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Oct 20, 2016 at 01:19:42AM -0700, Zhou Hao wrote:
> I think it is also necessary to validate these two fixed values in
> the Manifest

I agree.  But @stevvooe has been working to keep all intra-blob JSON
validation in image-spec [1], and you can perform this check in JSON
Schema.  I've just filed opencontainers/image-spec#404 with the JSON
Schema analog of this PR.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.log.html#l-71

stevvooe:
This validation belongs upstream.

PR #47, opened 2016-10-23
correct explanation in create runtime bundle

xiekeyang:
Signed-off-by: xiekeyang <keyang.xie@gmail.com>

### DIFF
--- a/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
+++ b/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
@@ -24,7 +24,7 @@ runtime-spec-compatible `dest/config.json`.
   A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
 
 **--type**
-  Type of the file to unpack. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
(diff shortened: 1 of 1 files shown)


cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
```diff
@@ -24,7 +24,7 @@ runtime-spec-compatible `dest/config.json`.
   A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
 
 **--type**
-  Type of the file to unpack. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
+  Type of the file to create runtime bundle. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
```
wking: I'm ok with the old language overloading “unpack” to cover both “unpack the rootfs” and “unpack the rootfs and translate the `config.json`”.  We cover the difference in the DESCRIPTION section above, so a bit of ambiguity here is fine.

If you can't accept the old language, I'd rather turn this around and focus on the fact that it's the input format (with the output being out-of-scope for this option).  Maybe:

> Source type. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout" or "image".

I'd also be ok with:

> `src` type. If unset…

xiekeyang: Yes, maybe it is redundant here to cover above description. And I'd prefer to old words than "source type".

PR #51, opened 2016-10-25
Add index.json validation

q384566678:
if the ManifestList exists, it needs to be validated

### DIFF
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -32,7 +32,7 @@ type bundleCmd struct {
 	typ      string // the type to bundle, can be empty string
 	ref      string
 	root     string
-	platform []string
+	platform string
 }
 
 func createHandle(context *cli.Context) error {
@@ -44,7 +44,7 @@ func createHandle(context *cli.Context) error {
 		typ:      context.String("type"),
 		ref:      context.String("ref"),
 		root:     context.String("rootfs"),
-		platform: context.StringSlice("platform"),
+		platform: context.String("platform"),
 	}
 
 	if v.typ == "" {
@@ -97,9 +97,9 @@ var createCommand = cli.Command{
 			Value: "rootfs",
 			Usage: "A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value.",
 		},
-		cli.StringSliceFlag{
+		cli.StringFlag{
 			Name:  "platform",
-			Usage: "The platform contains os and arch. Filter manifests according to the conditions provided. Only applicable if reftype is index.",
+			Usage: "Specify the os and architecture of the manifest, format is OS:Architecture. Only applicable if reftype is index.",
 		},
 	},
 }
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -29,9 +29,10 @@ var bundleTypes = []string{
 }
 
 type bundleCmd struct {
-	typ  string // the type to bundle, can be empty string
-	ref  string
-	root string
+	typ      string // the type to bundle, can be empty string
+	ref      string
+	root     string
+	platform []string
 }
 
 func createHandle(context *cli.Context) error {
@@ -40,9 +41,10 @@ func createHandle(context *cli.Context) error {
 	}
 
 	v := bundleCmd{
-		typ:  context.String("type"),
-		ref:  context.String("ref"),
-		root: context.String("rootfs"),
+		typ:      context.String("type"),
+		ref:      context.String("ref"),
+		root:     context.String("rootfs"),
+		platform: context.StringSlice("platform"),
 	}
 
 	if v.typ == "" {
@@ -56,10 +58,10 @@ func createHandle(context *cli.Context) error {
 	var err error
 	switch v.typ {
 	case image.TypeImageLayout:
-		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.ref, v.root)
+		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
 
 	case image.TypeImage:
-		err = image.CreateRuntimeBundleFile(context.Args()[0], context.Args()[1], v.ref, v.root)
+		err = image.CreateRuntimeBundleFile(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
 
 	default:
 		err = fmt.Errorf("cannot create %q", v.typ)
@@ -94,6 +96,10 @@ var createCommand = cli.Command{
 			Name:  "rootfs",
 			Value: "rootfs",
 			Usage: "A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value.",
+		},
+		cli.StringSliceFlag{
+			Name:  "platform",
+			Usage: "The platform contains os and arch. Filter manifests according to the conditions provided. Only applicable if reftype is index.",
 		},
 	},
 }
--- a/image/image.go
+++ b/image/image.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -140,12 +141,12 @@ func validate(w walker, refs []string, out *log.Logger) error {
 // UnpackLayout walks through the file tree given by src and, using the layers
 // specified in the manifest pointed to by the given ref, unpacks all layers in
 // the given destination directory or returns an error if the unpacking failed.
-func UnpackLayout(src, dest, ref string, platform []string) error {
+func UnpackLayout(src, dest, ref string, platform string) error {
 	return unpack(newPathWalker(src), dest, ref, platform)
 }
 
 // UnpackFile opens the file pointed by tarFileName and calls Unpack on it.
-func UnpackFile(tarFileName, dest, ref string, platform []string) error {
+func UnpackFile(tarFileName, dest, ref string, platform string) error {
 	f, err := os.Open(tarFileName)
 	if err != nil {
 		return errors.Wrap(err, "unable to open file")
@@ -159,11 +160,11 @@ func UnpackFile(tarFileName, dest, ref string, platform []string) error {
 // the manifest pointed to by the given ref, unpacks all layers in the given
 // destination directory or returns an error if the unpacking failed.
 // The destination will be created if it does not exist.
-func Unpack(r io.ReadSeeker, dest, refName string, platform []string) error {
+func Unpack(r io.ReadSeeker, dest, refName string, platform string) error {
 	return unpack(newTarWalker(r), dest, refName, platform)
 }
 
-func unpack(w walker, dest, refName string, platform []string) error {
+func unpack(w walker, dest, refName string, platform string) error {
 	if err := layoutValidate(w); err != nil {
 		return err
 	}
@@ -216,13 +217,13 @@ func unpack(w walker, dest, refName string, platform []string) error {
 // CreateRuntimeBundleLayout walks through the file tree given by src and
 // creates an OCI runtime bundle in the given destination dest
 // or returns an error if the unpacking failed.
-func CreateRuntimeBundleLayout(src, dest, ref, root string, platform []string) error {
+func CreateRuntimeBundleLayout(src, dest, ref, root string, platform string) error {
 	return createRuntimeBundle(newPathWalker(src), dest, ref, root, platform)
 }
 
 // CreateRuntimeBundleFile opens the file pointed by tarFile and calls
 // CreateRuntimeBundle.
-func CreateRuntimeBundleFile(tarFile, dest, ref, root string, platform []string) error {
+func CreateRuntimeBundleFile(tarFile, dest, ref, root string, platform string) error {
 	f, err := os.Open(tarFile)
 	if err != nil {
 		return errors.Wrap(err, "unable to open file")
@@ -235,11 +236,11 @@ func CreateRuntimeBundleFile(tarFile, dest, ref, root string, platform []string)
 // CreateRuntimeBundle walks through the given tar stream and
 // creates an OCI runtime bundle in the given destination dest
 // or returns an error if the unpacking failed.
-func CreateRuntimeBundle(r io.ReadSeeker, dest, ref, root string, platform []string) error {
+func CreateRuntimeBundle(r io.ReadSeeker, dest, ref, root string, platform string) error {
 	return createRuntimeBundle(newTarWalker(r), dest, ref, root, platform)
 }
 
-func createRuntimeBundle(w walker, dest, refName, rootfs string, platform []string) error {
+func createRuntimeBundle(w walker, dest, refName, rootfs string, platform string) error {
 	if err := layoutValidate(w); err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


vbatts:
LGTM
for a first pass

image/manifest_list.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		if err := checkPlatform(d); err != nil {
+			return errors.Wrap(err, "platform validation failed")
+		}
+	}
+
+	return nil
+}
+
+func checkPlatform(m Manifests) error {
```
wking: The consensus seems to be that this sort of intra-JSON-blob validation [should live][1] [in image-spec][2].

[1]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-254901311
[2]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-255001231

image/manifest_list.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ManifestDescriptor struct {
+	Descriptor v1.Descriptor
+
+	// Platform describes the platform which the image in the manifest runs on.
+	Platform Platform `json:"platform"`
+}
+
+func findManifest_list(w walker, d *descriptor) (*manifest_list, error) {
```
wking: Mixing camelCase and underscores is strange.  This should probably be `findManifestList`, although now that we have #29's `(r *tarReader) Get` and will hopefully soon have #5's CAS engine, I don't see much point to the `find*` functions.

wking:
The `Manifests` → `ManifestDescriptor` change has hidden my [earlier comment][0], but I think it's important enough to re-post with more visibility:

The consensus seems to be that this sort of intra-JSON-blob validation [should live][1] [in image-spec][2].  So I think `checkPlatform` belongs in image-spec, and not here.

[0]: https://github.com/opencontainers/image-tools/pull/51#discussion_r84950844
[1]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-254901311
[2]: https://github.com/opencontainers/image-tools/pull/39#issuecomment-255001231

coolljt0725:
@q384566678  build failed
````
image/image.go:98: cannot use err (type error) as type string in argument to strings.Contains
image/manifest_list.go:37: undefined: io in io.Reader
image/manifest_list.go:42: undefined: ioutil in ioutil.ReadAll
image/manifest_list.go:47: undefined: bytes in bytes.NewReader
image/manifest_list.go:72: d.Descriptor.validate undefined (type "github.com/opencontainers/image-tools/vendor/github.com/opencontainers/image-spec/specs-go".Descriptor has no field or method validate)
````

**Comment by q384566678** [comment]
I know,I'm trying to fix that @coolljt0725

File: image/manifest_list.go [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+	case errEOW:
+		return &ml, nil
+	default:
+		return nil, err
+	}
+}
+
+func (ml *manifestlist) validate(w walker) error {
+	for _, d := range ml.Manifests {
+		if err := d.Descriptor.validate(w, []string{v1.MediaTypeImageConfig}); err != nil {
```
q384566678: But the manifest also have  [spec's Descriptor](https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L25), it can use [validate](https://github.com/opencontainers/image-tools/blob/master/image/manifest.go#L78) ,I think as long as the manifest file to the Descriptor as specific, you can call.

q384566678: But the manifest also have  [spec's Descriptor](https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L25), it can use [validate](https://github.com/opencontainers/image-tools/blob/master/image/manifest.go#L78) ,I think as long as the manifest file to the Descriptor as specific, you can call.

wking:
On Mon, Oct 31, 2016 at 01:54:49AM -0700, Zhou Hao wrote:
> +		if err := d.Descriptor.validate(w, []string{v1.MediaTypeImageConfig}); err != nil {
> 
> But the manifest also have [spec's
> Descriptor](https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L25),
> it can use
> [validate](https://github.com/opencontainers/image-tools/blob/master/image/manifest.go#L78)
> ,I think as long as the manifest file to the Descriptor as specific,
> you can call.

That validate call is using the image-tools-specific manifest [1],
which uses the image-tools-specific descriptor [2].  I don't think
image-tools-specific types are a good pattern to follow, but you'll
need to convert v1.Descriptors to image-tools-specific descriptors
before you can call the image-tools-specific-descriptor's Validate
[3].  Ideally these would all be functions that took the image-spec
types.

[1]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/manifest.go#L77
[2]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/manifest.go#L36
[3]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/descriptor.go#L97

image/image.go
```diff
@@ -87,6 +88,18 @@ func validate(w walker, refs []string, out *log.Logger) error {
...
 			return err
 		}
+
+		ml, err := findManifestList(w, d)
+		if err == nil {
+			if err := ml.validate(w); err != nil {
+				return err
+			}
+		} else {
+			if !strings.Contains(string(err), "manifestlist not found") {
```
coolljt0725: At the moment, the manifest list is OPTIONAL(https://github.com/opencontainers/image-spec/blob/master/manifest-list.md#oci-image-manifest-list-specification) , so we should not error out if there is no manifest list

image/image.go
```diff
@@ -87,6 +88,18 @@ func validate(w walker, refs []string, out *log.Logger) error {
...
+				return err
+			}
+		}
+
 		if out != nil {
```
coolljt0725: Should we `findManifestList` first and if manifest_list exists, then extract the manifest from the manifest_list and then validate the manifest?

I +1 with @wking 's proposal 
>I'd rather land opencontainers/image-spec#403 and use:

>Get an initial descriptor.
Fetch the object referenced by the descriptor.
Run the appropriate intra-blob validation on it (opencontainers/image-spec#403).
For any children referenced from the blob, recurse to (2).

image/image.go
q384566678: Yes, I was in accordance with the idea to write, I first made a judgment on the error message, if it is because there is no manifest lis and return, not error out.（There is one! The symbol follows the if）

image/image.go
```diff
@@ -87,6 +87,16 @@ func validate(w walker, refs []string, out *log.Logger) error {
 		if err := m.validate(w); err != nil {
 			return err
 		}
+
+		ml, err := findManifestList(w, d)
```
coolljt0725: This seems not correct, `findManifestList(w, d)` here just assume the `d` is a `manifest list`, but actually, `d` could be manifest or manifest list.   This is why the test failed.
````
--- FAIL: TestValidateLayout (0.03s)
	image_test.go:175: blobs/sha256/3b6be0edf862ac9e6e790fb010cf26770d9ab266f3ff81b5b82f0cf8d6f7d1a6: manifestlist validation failed: [manifests: manifests is required]
````
I think we should check the `d` after [`d.validate(w, validRefMediaTypes)`] (https://github.com/opencontainers/image-tools/pull/51/files#diff-da07f373046a0a24e67b18699dc15a70L78), if it is a manifest, then go to `findManifest`, if it is a manifest list, then go to `findManifestList`

Comment (q384566678) [inline image/image.go]: Nice,I will do that.Thank you for your suggestion.

**Comment by coolljt0725** [comment]
test failed 
````
go test -race -cover github.com/opencontainers/image-tools/cmd/oci-create-runtime-bundle github.com/opencontainers/image-tools/cmd/oci-image-validate github.com/opencontainers/image-tools/cmd/oci-unpack github.com/opencontainers/image-tools/image
?   	github.com/opencontainers/image-tools/cmd/oci-create-runtime-bundle	[no test files]
?   	github.com/opencontainers/image-tools/cmd/oci-image-validate	[no test files]
?   	github.com/opencontainers/image-tools/cmd/oci-unpack	[no test files]
--- FAIL: TestValidateLayout (0.03s)
	image_test.go:175: blobs/sha256/f175ad817b1c75958e74e224f57b17d3a3e746bc434737b1bde837dbe7d33409: manifest validation failed: [config: config is required layers: layers is required]
FAIL
coverage: 29.3% of statements
FAIL	github.com/opencontainers/image-tools/image	0.096s
make: *** [test] Error 1
````

**Comment by q384566678** [comment]
This is my idea about the manfestlist related validation. PTAL @vbatts @wking @coolljt0725 @stevvooe @xiekeyang

**Comment by xiekeyang** [comment]
> The consensus seems to be that this sort of intra-JSON-blob validation should live in image-spec. So I think checkPlatform belongs in image-spec, and not here.

I agree with @wking. I think the job of finding CAS blob belongs to image-tools, and JSON blob validation belongs to image-spec. As to this commit, it (of course and previous framework) mix them. So we might had better to refactor this according to our consensus firstly.

**Comment by wking** [comment]
On Mon, Dec 05, 2016 at 11:27:33PM -0800, xiekeyang wrote:
> So we might had better to refactor this according to our consensus
> firstly.

My preferred approach to this is with a media-type callback registry.
opencontainers/image-spec#403 is floating callback-based validation.
I'd like to see image-tools providing callback-based unpacking:

  var Unpackers = map[string]Unpacker{
    v1.MediaTypeImageManifestList:          UnpackManifestList,
    v1.MediaTypeImageManifest:              UnpackManifest,
    v1.MediaTypeImageConfig:                UnpackConfig,
    v1.MediaTypeImageLayer:                 UnpackLayer,
    v1.MediaTypeImageLayerNonDistributable: UnpackLayer,
  }

I'd also like to see image-tools providing validation which combined
image-spec-provided intra-JSON-blob validation with code to discover
and validate referenced ancestors (see @cyphar's clever reflect-based
child walking [1] for a possible approach to handling this without
adding per-type image-tools validation code).

[1]: https://github.com/cyphar/umoci/blob/master/image/cas/gc.go#L39-L120

**Comment by q384566678** [comment]
At present according to my method is findmanifest to write, in oci - layout, likely is the manifest, it could also be manifestlist, if manifestlist, will be based on manifestlist to find again after get the manifest to verify.

File: image/image.go [hunk 7]
```diff
@@ -166,15 +231,54 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
...
+			if err != nil {
+				return err
+			}
+
+			if err := m.validate(w); err != nil {
+				return err
+			}
+
+			if err := createRuntimebundle(w, m, dest, rootfs); err != nil {
+				return err
```
coolljt0725: On creating runtime bundle, I don't think we should do this directly, because the manifest list contains different manifest of architecture or platform, we should not unpack all the manifest to a same `dest`. I think we should add a flag to `oci-create-runtime-bundle` and `oci-unpack`  to specify the arch we want to unpack or the `oci-create-runtime-bundle` and `oci-unpack` detect the target arch and unpack the target arch manifest.

coolljt0725:
And please add tests

image/image.go
q384566678: If you want to increase the flag, first of all to know which contains which arch, but these do not know in advance.

q384566678: Is not able to contain all the manifest are listed, let the user choose which one to use.

cyphar:
It would be very nice if we could switch this to use `umoci`'s blob introspection code. I'm going to start pushing code from `umoci` into this repo soon.

q384566678:
@cyphar @stevvooe @vbatts @coolljt0725 @wking I have now updated the validation of the index, and there is a problem: which manifest in index should be used when the ref finds index? Is it necessary to additional input parameters to determine, or is there any other way?

xiekeyang:
@q384566678 Maybe all manifests in index should be validated. One of index application is that it included manifests under different platform. For this case, all manifests are validated and checked the platform-matched one. It is my opinion.

q384566678:
@xiekeyang I agree that you validate all manifest views. But when the line `create` and `unpack` programs need accurate platform parameters, I'm not sure whether to enter this parameter or to execute all the manifest.

cmd/oci-image-tool/create.go
```diff
@@ -40,9 +41,10 @@ func createHandle(context *cli.Context) error {
 	}
 
 	v := bundleCmd{
-		typ:  context.String("type"),
-		ref:  context.String("ref"),
-		root: context.String("rootfs"),
+		typ:      context.String("type"),
+		ref:      context.String("ref"),
+		root:     context.String("rootfs"),
+		platform: context.StringSlice("platform"),
```
stevvooe: What is the expected usage of this flag?

q384566678: As I said above, when you find and validate multiple `manifest` in `index`, the following create and unpack functions don't know which `manifest` is called specifically. I added this flag to add input parameters to determine the specific `manifest`.

stevvooe: So, don't explain it to me: document the flag in the code and help functions. Based on this, I have no clue how to use it. Platform is a nested structure and that needs to represented as a filter here.

q384566678: I made the following changes:
1. Added a filter function to filter the manifests
2. Added a more specific description of the platform flag, easy to use

PTAL, What other questions please put forward. Thanks.

q384566678:
@vbatts @coolljt0725 @cyphar @xiekeyang @stevvooe @jonboulle I updated test file, PTAL.

cmd/oci-image-tool/create.go
```diff
@@ -95,5 +97,9 @@ var createCommand = cli.Command{
 			Value: "rootfs",
 			Usage: "A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value.",
 		},
+		cli.StringSliceFlag{
+			Name:  "platform",
+			Usage: "The platform contains os and arch. Filter manifests according to the conditions provided. Only applicable if reftype is index.",
```
stevvooe: I still don't know how to use this from this description. Is there an example?

man/oci-image-tool-unpack.1.md
```diff
@@ -20,6 +20,11 @@ oci-image-tool unpack \- Unpack an image or image source layout
 **--type**=""
   Type of the file to unpack. If unset, oci-image-tool will try to auto-detect the type. One of "imageLayout,image"
 
+**--platform**=""
+  Specify the os and architecture of the manifest, format is OS:Architecture.
+  e.g. --platform linux:amd64
```
q384566678: @stevvooe I have modified the introduction of information and added examples. PTAL

q384566678:
reping @opencontainers/image-tools-maintainers

cyphar:
Just to make sure, if I want to validate an entire image with `./oci-image-tool validate --type imageLayout` it will still work? In other words, is the `--platform` filter the same as the `--ref` filter in that they don't filter anything if you don't specify them?

image/image.go
```diff
@@ -228,3 +311,37 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
...
+// filertManifest returns a filtered list of manifests
+func filterManifest(w walker, Manifests []v1.Descriptor, platform []string) ([]*manifest, error) {
+	var manifests []*manifest
+
+	if len(Manifests) == 0 {
+		fmt.Println("warning: no manifests found")
+		return manifests, nil
+	}
+
+	if len(platform) != 2 {
```
cyphar: Either this should be checked far earlier (when you do the splitting) or you should only split it here. I would prefer if you just passed the string down to `filterManifest` and then you verify the arguments here.

q384566678:
> Just to make sure, if I want to validate an entire image with ./oci-image-tool validate --type imageLayout it will still work? 

Yes.

> In other words, is the --platform filter the same as the --ref filter in that they don't filter anything if you don't specify them?

Yes, just to filter the manifest in certain circumstances.

image/image.go
q384566678: updated.

cyphar:
LGTM.

q384566678:

xiekeyang:
LGTM

It is OK from my side, PTAL from others.

q384566678:
If someone else has no objection, I will merge it tomorrow.

q384566678:
@cyphar @xiekeyang @coolljt0725  rebased, PTAL, thanks

q384566678:
reping @vbatts @coolljt0725 @xiekeyang @cyphar @stevvooe PTAL

xiekeyang:
LGTM

cyphar:
LGTM.

PR #61, opened 2016-10-28
Add regular expression match for ref string input

TrumanLing:
implementation for spec description "Object names in the refs
subdirectories MUST NOT include characters outside of the set of "A" to
"Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore
_." at https://github.com/opencontainers/image-spec/blob/master/image-layout.md.

Signed-off-by: Ling FaKe lingfake@huawei.com

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -66,6 +66,12 @@ func validate(w walker, refs []string, out *log.Logger) error {
 		}
 	}
 
+	for _, refName := range refs {
+		if err = refsNameRegexpCheck(refName); err != nil {
+			return err
+		}
+	}
+
 	for _, ref := range refs {
 		d, ok := ds[ref]
 		if !ok {
@@ -115,6 +121,10 @@ func Unpack(tarFile, dest, ref string) error {
 }
 
 func unpack(w walker, dest, refName string) error {
+	if err := refsNameRegexpCheck(refName); err != nil {
+		return err
+	}
+
 	ref, err := findDescriptor(w, refName)
 	if err != nil {
 		return err
@@ -157,6 +167,10 @@ func CreateRuntimeBundle(tarFile, dest, ref, root string) error {
 }
 
 func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
+	if err := refsNameRegexpCheck(refName); err != nil {
+		return err
+	}
+
 	ref, err := findDescriptor(w, refName)
 	if err != nil {
 		return err
@@ -197,4 +211,11 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
 	defer f.Close()
 
 	return json.NewEncoder(f).Encode(spec)
+}
+
+func refsNameRegexpCheck(refName string) error {
+	if boolRefsNameMatch := v1.RefsRegexp.MatchString(refName); !boolRefsNameMatch {
+		return fmt.Errorf("input parameter --ref (%s) is invalide", refName)
+	}
+	return nil
 }


cmd/oci-create-runtime-bundle/main.go
```diff
@@ -29,6 +30,9 @@ import (
 // and will be populated by the Makefile
 var gitCommit = ""
 
+// refRegexp for string ref should be complied with description "Object names in the refs subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore _." at https://github.com/opencontainers/image-spec/blob/master/image-layout.md.
+var refRegexp = regexp.MustCompile(`^[a-zA-Z0-9\-\.\_]+?$`)
```
wking: This constant should probably be provided by image-spec, from which runtime-tools can import and use it.

cmd/oci-create-runtime-bundle/main.go
```diff
@@ -114,6 +118,18 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
...
+	stringRefFind := refRegexp.FindString(v.ref)
+	v.stderr.Printf("ref name is %s", stringRefFind)
+
+	//check whether string ref match regular expression or not
+	boolRefMatch := refRegexp.MatchString(v.ref)
+	v.stderr.Printf("matching result is %v", boolRefMatch)
+	if boolRefMatch != true {
+		v.stderr.Printf("ref name (%s) is invalid", v.ref)
+		os.Exit(1)
+	}
```
wking: This probably belongs in image-layout validation like `image/image.go`'s `validate`.

wking: There is already [ref-oriented validation in `image/image.go`'s `validate/][1].  Adding that check here saves a few cycles in the `oci-create-runtime-bundle` case, but completely misses the validation check in the `oci-image-validate` case and the “library call to `ValidateLayout`” case.

[1]: https://github.com/TrumanLing/image-tools/blob/adbe39359bf2aac1aa01d3ce376036039c19fe32/image/image.go#L69

TrumanLing: @wking Greate suggestion! But `image/image.go's validate` is for `oci-image-validate`, so I think, it is reasonable to add `regular expression` checks at the beginning of `image/image.go's` `unpack`, `validate` and `createRuntimeBundle`, which are separately for `oci-unpack`, `oci-image-validate` and `oci-create-runtime-bundle`.

image/image.go
```diff
@@ -49,6 +49,13 @@ var validRefMediaTypes = []string{
 }
 
 func validate(w walker, refs []string, out *log.Logger) error {
+	var tmperr error
```
wking: No need for `tmperr`.  I'd rather have:

```go
func validate(w walker, refs []string, out *log.Logger) (err error) {
```

above, use `err` for `refsNameRegexpCheck`, and possibly shift an `:=` → `=` below where we currently setup `err`.

image/image.go
```diff
@@ -115,6 +122,10 @@ func Unpack(tarFile, dest, ref string) error {
 }
 
 func unpack(w walker, dest, refName string) error {
+	if err := refsNameRegexpCheck(refName); err != nil {
```
wking: I'm not convinced that we want to be doing validation as part of unpacking (folks who want it can call the validation code *before* calling `unpack`).  But if that is not convincing, we can always add this for now, and remove it if/when someone gets around to filing a PR to pull the existing validation logic out of the unpacking code.

image/image.go
TrumanLing: IMHO, the regular expression check should be at the beginning of the function. As for your recommendation, let the check below line `:=` for `err` to avoid one more variable. But I think it is a little bit of breaking the logic with handling image validation, that means, the original func `validate` is for image validation, and not for parameter input check. So, I think the parameter check should be at the beginning, and the rest of code lines does its original operation, no insertion of other operation, even just only one line.

image/image.go
TrumanLing: >I'm not convinced that we want to be doing validation as part of unpacking (folks who want it can call the validation code before calling unpack).

image/image.go
wking: > IMHO, the regular expression check should be at the beginning of the function.

I hadn't thought about that, but not that you bring it up… ;).  I think it should be [here][1], because:

1. If you run your check too early, you might not check anything and have invalid refs [injected here][2] slip through without getting caught.
2. A ref [not existing][3] seems more important than not meeting the regexp.

I feel pretty strongly about (1), but not particularly strongly about (2).  Regardless of how (2) goes, the check should be inside that loop.  Perhaps the point that the validated refs may not be parameters (when they come from `listReferences`) is enough to allow closer integration between this regexp validation and the existing logic.

[1]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/image.go#L77
[2]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/image.go#L63-L65
[3]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/image.go#L75

TrumanLing: I agree with you, but I updated it [here](https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/image.go#L68), because `ref` is used at [here](https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/image/image.go#L69)

image/image.go
wking: On Sun, Oct 30, 2016 at 07:11:09PM -0700, TrumanLing wrote:
> … but we can't avoid user's typing error…

That's not a problem.  If the user typo's the ref name, unpacking will fail because the ref does not exist.  Checking against the regexp in that case only “protects” you from the unlikely and presumably intentional case where the user is trying to unpack an illegal ref *that already exists in the ref store*.  If the illegal ref is in the ref store it's not a compliant image, but I still think we should allow the user to unpack the ref (assuming the DAG it points to is unpackable).

TrumanLing: OK, I understand your concern. I will change it after [PR](https://github.com/opencontainers/image-spec/pull/421) in image-spec is merged, which leads to CI failure without merging it.

vbatts:
failures in `make lint`

wking:
On Wed, Feb 08, 2017 at 11:29:27AM -0800, Vincent Batts wrote:
> failures in `make lint`

Is it worth landing this for rc4 support with refs (and this
restriction on their names) now removed from the spec for the next
release [1] and no similar restriction on org.opencontainers.ref.name
values [2]?  I think we should just close this PR and wait for someone
to file one with a more comprehensive catch-up to
opencontainers/image-spec#533.

[1]: https://github.com/opencontainers/image-spec/pull/533/files#diff-99bfcb610f238b942638d49d1ff2eaf4L58
[2]: https://github.com/opencontainers/image-spec/pull/533/files#diff-338aacf6337d00fc0cde843c3182c3f8R24

xiekeyang:
Close for https://github.com/opencontainers/image-spec/pull/533

PR #62, opened 2016-10-31
Modify the automatic determination of the Config type

q384566678:
The Config attribute may be empty, so cannot be used as the basis for judging the Config type, should use the RootFS is inevitable in this attribute to judge .

### DIFF
--- a/image/autodetect.go
+++ b/image/autodetect.go
@@ -79,8 +79,8 @@ func Autodetect(path string) (string, error) {
 
 	header := struct {
 		SchemaVersion int         `json:"schemaVersion"`
-		MediaType     string      `json:"mediaType"`
 		Config        interface{} `json:"config"`
+		Manifests     interface{} `json:"manifests"`
 	}{}
 
 	if err := json.NewDecoder(f).Decode(&header); err != nil {
@@ -97,13 +97,13 @@ func Autodetect(path string) (string, error) {
 	}
 
 	switch {
-	case header.MediaType == string(schema.ValidatorMediaTypeManifest):
+	case header.SchemaVersion == 2 && header.Config != nil:
 		return TypeManifest, nil
 
-	case header.MediaType == string(schema.ValidatorMediaTypeImageIndex):
+	case header.Manifests != nil:
 		return TypeImageIndex, nil
 
-	case header.MediaType == "" && header.SchemaVersion == 0 && header.Config != nil:
+	case header.SchemaVersion == 0 && header.Config != nil:
 		// config files don't have mediaType/schemaVersion header
 		return TypeConfig, nil
 	}


q384566678:

q384566678:
ping @vbatts @wking @xiekeyang @coolljt0725  PTAL

wking:
I have no opinion on peek-inside type detection implementations ;).

q384566678:
PTAL @stevvooe @philips @jbouzane @jstarks

image/autodetect.go
```diff
@@ -103,8 +102,8 @@ func Autodetect(path string) (string, error) {
 	case header.MediaType == string(schema.MediaTypeManifestList):
 		return TypeManifestList, nil
 
-	case header.MediaType == "" && header.SchemaVersion == 0 && header.Config != nil:
-		// config files don't have mediaType/schemaVersion header
+	case header.RootFS != nil:
```
xiekeyang: I think it is good to check `rootfs`(required) instead `config`(optional).
However, statement `header.MediaType == "" && header.SchemaVersion == 0` had better to be kept, which are also terms differ with other json descriptors.

q384566678: I think only need a unique attribute can determine the Config type, the other on some superfluous

xiekeyang:
As to this PR, I think it is already good. However, @stevvooe suggested that validation work for JSON files belongs to image-spec repo, which seems to include Auto-detect on JOSN file.

@stevvooe WDYT? This modification is OK, or should remove to image-spec repo?

stevvooe:
@xiekeyang I am not sure what this is trying to do. Unfortunately, the use of mediaType to detect the type doesn't work, since descriptors describe the target to which they are pointing.

Really, trying to ascertain what a thing is with having a qualified reference presents a host of security problems. Really, you want to instruct the process what the type of the target item is dispatch to a handler. The type resolution is a trusted, deterministic process, either by assuming that all incoming objects are of the same type (such as descriptors) or through a secured mediatype (such as a signed metadata provider).

If we really want content auto detection, we need to have a hard look at the spec, but I don't think is really needed in any real use case.

q384566678:
I think there may be JSON file here to verify the location of the problem, but this method of verification I think it is possible, I have already explained above

xiekeyang:
@stevvooe 
I think you are encouraging to remove JSON media detection from image tools, as it is unnecessary, not deterministic and insecure. Right?
I agree that, because current detect use its media and specific object, which really seems not precise. If it is needed, at least it should be done in image spec repo.

stevvooe:
@xiekeyang Correct.

@vbatts Do we have an open issue for this?

q384566678:
Means the JSON related validation into image - spec?

xiekeyang:
> Means the JSON related validation into image - spec?

I think so. And media type auto-detection might had better to be removed.

q384566678:
https://github.com/opencontainers/image-spec/pull/481

q384566678:
@xiekeyang @stevvooe  What do you think of this change?

image/autodetect.go
```diff
@@ -67,46 +66,11 @@ func Autodetect(path string) (string, error) {
 		return TypeImage, nil
 
 	case "text/plain; charset=utf-8":
```
stevvooe: Where is `charset` coming from? This looks really wrong.

q384566678: Uh，This is the original code, the results by `http.DetectContentType` to determine the format.

stevvooe: Is there any reason autodetect can't be the default? This particular media type doesn't make sense. Media types don't actually have charsets (that is content types).

q384566678:
@xiekeyang  @coolljt0725  On this PR, I think I can do is make such a change.PTAL

xiekeyang:
Detection/validation stuff might need a issue involving design spec firstly. Based on which we should get the agreement. And then start to implement.

vbatts:
Is this still needed?

xiekeyang:
Closed. 
@q384566678 Could you or others please put up an issue, to discuss about the framework of validation/detection/checking functionality on mediatype, json schema, length of array ...

q384566678:
@xiekeyang @stevvooe @vbatts I think if we can reopen this PR? Because the current verification method `Autodetect` has fallen behind (the `mediatype` is not necessary) .Or move to imag-spec to verify?

q384566678:
I reopen this pr because the current `auduto` is wrong, `mediatype` has become optional.
ping @opencontainers/image-tools-maintainers PTAL.

stevvooe:
@q384566678 Where does this get used? The autodetection looks right but it will be naive to the introduction of new types.

q384566678:
> {
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 622,
      "digest": "sha256:cd4c216fcb2ce4f1e1c18f2d53dd5a00f72b27ae21eaccafabe4c00b86e9326e",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 622,
      "digest": "sha256:cd4c216fcb2ce4f1e1c18f2d53dd5a00f72b27ae21eaccafabe4c00b86e9326e",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ],
  "annotations": {
    "com.example.index.revision": "r124356"
  }
}

@stevvooe  Above is an example of an `index` type, if we execute the `validate` command on it will fail:

> ➜  image-tools git:(master) ./oci-image-tool validate ../index 
1 errors detected: 
../index: validation failed: unable to determine type: unknown media type

Because through `Autodetect` can not correctly determine its type.

stevvooe:
@q384566678 But why are we using `Autodect` here? Isn't this not an `index.json` where we already know the type?

q384566678:
@stevvooe  Because we entered `./oci-image-tool validate ../index`  instead of `./oci-image-tool validate --type imageIndex ../index` . If you enter a type with  `--type `, you do not need to execute the `Autodect` function, but if you do not enter the validation type, you must use the `Autodect` function to determine the type.
The following is the relevant code：
```
	if typ == "" {
		if typ, err = image.Autodetect(name); err != nil {
			return errors.Wrap(err, "unable to determine type")
		}
	}
```

stevvooe:
@q384566678 What is the use case of autodetect? This seems like the wrong direction.

q384566678:
@stevvooe The function of this function is when you execute the command, if not enter type, to determine the type. And there is no problem.

stevvooe:
@q384566678 What? This seems incorrect. How can you validate something if you don't know the type?

q384566678:
@stevvooe Uh, as I said. if you don't enter the type (or you don't know the type), the `Autodect` will determine the type according to the path.

q384566678:
@xiekeyang @coolljt0725 PTAL

q384566678:
@xiekeyang @stevvooe @coolljt0725 @cyphar I think this problem needs to have a result, it's been a long time. I still think it's necessary to do this, or there will be validation errors like I said above.

xiekeyang:
I still feel that we should follow [1] and [2].

[1]https://github.com/opencontainers/image-tools/pull/62#issuecomment-263528341
[2]https://github.com/opencontainers/image-tools/pull/62#issuecomment-264027138

stevvooe:
@q384566678 I am not sure how to make this more clear: autodetection of types in OCI images is not a supported concept. It is insecure and broken in practice (see the history of the tar format). If we keep trying to do this, without major design changes, we will have the same broken behavior.

Let's remove autodetection and find ways in which we can accurately and correctly resolve types.

q384566678:
@stevvooe ok, I'll do it.

q384566678:
> It is insecure and broken in practice (see the history of the tar format)

@stevvooe Do you have a specific link to this ? I can not find it.Thanks.

q384566678:
@stevvooe Can you explain the above question?

q384566678:
ping @stevvooe

q384566678:
ping @stevvooe @coolljt0725 @xiekeyang How about this revision?

cmd/oci-image-tool/validate.go
```diff
@@ -121,6 +121,12 @@ func validatePath(name string) error {
 		return schema.ValidatorMediaTypeImageIndex.Validate(f)
 	case image.TypeConfig:
 		return schema.ValidatorMediaTypeImageConfig.Validate(f)
+	case "JSON":
```
coolljt0725: I don't think we introduce a new type `JSON` is a good idea here, `JSON` is not any type in image spec. I think it's better to keep the old way.

q384566678: I also think the previous method is good, and I have not thought of other methods at present, but the suggestion of @stevvooe  is to replace the previous method.

Mashimiao:
I think we should close this PR.
Though we may have different opinions on keep autodetect for image or not in #184,
but we all agree that we should remove autodetect for image files, Right?
So, I think we don't need this PR any more.
@vbatts  @stevvooe  @xiekeyang  @coolljt0725

stevvooe:
@Mashimiao I think we can close this.

PR #64, opened 2016-10-31
Increase unpack type error judgment

q384566678:
An error should be reported when the unpack type is not one of the specified two types

### DIFF
--- a/cmd/oci-create-runtime-bundle/main.go
+++ b/cmd/oci-create-runtime-bundle/main.go
@@ -130,6 +130,10 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 
 	case image.TypeImage:
 		err = image.CreateRuntimeBundle(args[0], args[1], v.ref, v.root)
+
+	default:
+		err = fmt.Errorf("cannot create %q", v.typ)
+
 	}
 
 	if err != nil {
--- a/cmd/oci-unpack/main.go
+++ b/cmd/oci-unpack/main.go
@@ -117,6 +117,9 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 
 	case image.TypeImage:
 		err = image.Unpack(args[0], args[1], v.ref)
+
+	default:
+		err = fmt.Errorf("cannot unpack %q", v.typ)
 	}
 
 	if err != nil {


cmd/oci-unpack/main.go
```diff
@@ -117,6 +118,9 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 
 	case image.TypeImage:
 		err = image.Unpack(args[0], args[1], v.ref)
+
+	default:
+		err = errors.New("Error media type")
```
wking: I'd rather list the type.  How about:

```go
err = fmt.Errorf("cannot unpack %q", v.typ)
```

wking:
On Mon, Oct 31, 2016 at 12:21:42AM -0700, Zhou Hao wrote:
> An error should be reported when the unpack type is not one of the
> specified two types.

I don't think the spec backs you up on this.  It has [1]:

   Commonly pointing to an image manifest or an image manifest list.

not:

   MUST reference an image manifest or an image manifest list.

However, there are clearly types that image-tools doesn't know how to
unpack (e.g. single layers).  And some types that image-tools will
never know how to unpack (e.g. image/png).  So I'm in favor of this
warning, but think it is based on image-tools support and not on spec
limitations.

[1]: https://github.com/opencontainers/image-spec/blame/5dbe4a342b6de21149f5c9f23c13b62ec1503082/image-layout.md#L15

cmd/oci-unpack/main.go
q384566678: Yes, it would be better to do so, thank you

q384566678:
> I don't think the spec backs you up on this. It has [1]: Commonly pointing to an image manifest or an image manifest list.

On the other hand, the value of type can also be imported, so it is more necessary to add judgment

cmd/oci-unpack/main.go
xiekeyang: Here seems better to refer to https://github.com/opencontainers/image-tools/blob/master/cmd/oci-image-validate/main.go#L176, as fmt.Errorf("type %q unimplemented", v.typ) , and exit by `os.Exit(1)`.

wking: I think it's good to say what it is that's not implemented (unpacking in this case) to distinguish among all the things that might not be implemented.

q384566678:
I wonder if this community is very inactive? Why do I have those PR piling up in there, not been merged, do not know if it should be closed, so that my work has become very troubled @wking @xiekeyang @coolljt0725 @vbatts @stevvooe @philips

stevvooe:
LGTM

@q384566678 It is very hard for the maintainers to comment with the volume of messages generated by @wking. This is compounded by the fact that much of what he is saying is either wildly wrong or incoherent.

In this case, the specification has no bearing on whether the tool can unpack the image. If the tool doesn't recognize the type, it should throw an error.

cmd/oci-create-runtime-bundle/main.go
```diff
@@ -130,6 +130,10 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 
 	case image.TypeImage:
 		err = image.CreateRuntimeBundle(args[0], args[1], v.ref, v.root)
+
+	default:
+		err = fmt.Errorf("cannot create %q", v.typ)
```
xiekeyang: If it is better to use `%s` here, because `v.typ` is string type?

cmd/oci-unpack/main.go
```diff
@@ -117,6 +117,9 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 
 	case image.TypeImage:
 		err = image.Unpack(args[0], args[1], v.ref)
+
+	default:
+		err = fmt.Errorf("cannot unpack %q", v.typ)
```
xiekeyang: If it is better to use `%s` here, because `v.typ` is string type?

xiekeyang:
@q384566678 I want to merge all the approved prs, but unfortunately I still have no permission for `green approve`.

q384566678:
@xiekeyang This is the case, I do not understand these, I just want my PR to have a faster result, so I @ the managers, if there is an offense, please forgive me

xiekeyang:
LGTM

cmd/oci-create-runtime-bundle/main.go
wking: Using `%q` gives you unambiguous quoting when the type contains strange characters:

```
$ cat test.go 
package main

import "fmt"

func main() {
        fmt.Printf("%q\n", "foo\"bar")
        fmt.Printf("%s\n", "foo\"bar")
        fmt.Printf("%q\n", "foo\nbar")
        fmt.Printf("%s\n", "foo\nbar")
}
$ go run test.go 
"foo\"bar"
foo"bar
"foo\nbar"
foo
bar
```

In this case, `v.typ` can be set directly by the user, so I think we should avoid assuming that it contains no strange characters.

q384566678: For the two I feel are almost, but according to the above test and [other uses](https://github.com/opencontainers/image-tools/blame/master/cmd/oci-image-validate/main.go#L176) ， %q is more appropriate point

stevvooe:
LGTM

coolljt0725:
LGTM

xiekeyang:
LGTM

xiekeyang:
LGTM

PR #65, opened 2016-10-31
descriptor: size verification should be before digest calculation

TrumanLing:
Conforming to description 'Before calculating the digest, the size of the content SHOULD be
verified to reduce hash collision space.' at https://github.com/opencontainers/image-spec/blob/master/descriptor.md.

Signed-off-by: Ling FaKe <lingfake@huawei.com>

### DIFF
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -122,14 +122,14 @@ func (d *descriptor) validateContent(r io.Reader) error {
 		return errors.Wrap(err, "error generating hash")
 	}
 
+	if n != d.Size {
+		return errors.New("size mismatch")
+	}
+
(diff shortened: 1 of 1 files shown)


wking:
0d2a6069 looks good to me, although with the Copy approach currently
used by validateContent there's not much benefit in either order.  The
spec wording *does* increase efficiency in cases where you can
calculate the size more cheaply (e.g. if you're reading a file off the
disk and can stat for the size, or if you're recieving a file over
HTTP and can look at Content-Length).

coolljt0725:
LGTM

vbatts:
LGTM

PR #67, opened 2016-11-01
omit error for unimplemented unpacking types

xiekeyang:
For unimplemented(so far) unpacking types, it should omit error message, and return failed result.


wking:
Dup of #64?

xiekeyang:
yes, thanks.

PR #73, opened 2016-11-02
autodetect.go: combine two switch cases

TrumanLing:
Signed-off-by: Ling FaKe lingfake@huawei.com

### DIFF
--- a/image/autodetect.go
+++ b/image/autodetect.go
@@ -60,10 +60,7 @@ func Autodetect(path string) (string, error) {
 	mimeType := http.DetectContentType(buf)
 
 	switch mimeType {
-	case "application/x-gzip":
-		return TypeImage, nil
-
-	case "application/octet-stream":
+	case "application/x-gzip", "application/octet-stream":
 		return TypeImage, nil
 
 	case "text/plain; charset=utf-8":


image/autodetect.go
```diff
@@ -61,8 +61,6 @@ func Autodetect(path string) (string, error) {
 
 	switch mimeType {
 	case "application/x-gzip":
-		return TypeImage, nil
-
 	case "application/octet-stream":
```
wking: I don't think Go's `switch` supports fall-through without an [explict `fallthrough`][1].

[1]: https://golang.org/ref/spec#Fallthrough_statements

TrumanLing: @wking It is no problem, I have verified it.

In generally, Go's `switch` supports style `case <con1>,<con2>:` and that I have submitted.

wking: `case "application/x-gzip", "application/octet-stream":` would be fine, but what you have here is:

```go
case "application/x-gzip":
case "application/octet-stream":
```

And testing a similar layout locally:

```
$ cat x.go 
package main
import ("fmt")
func main() {
        var a = 0
        switch a {
        case 0:
        case 1:
                fmt.Println("hi")
        }
}
$ go run x.go 
```

I don't see `hi` in my stdout.

TrumanLing:
@wking updated it, PTAL.

wking:
8cc35783d looks good to me.

coolljt0725:
LGTM

vbatts:
LGTM

PR #77, opened 2016-11-06
image: generate useful config

### LINKED ISSUE #76: create-runtime-bundle: generates incorrect Linux config
Currently if you take a random Docker image's config file (then translate it to an OCI one using  `skopeo`) you'll get a `config.json` which looks like this:

```
{
  "ociVersion": "1.0.0-rc2",
  "platform": {
    "os": "linux",
    "arch": "amd64"
  },
  "process": {
    "terminal": true,
    "user": {
      "uid": 0,
      "gid": 0
    },
    "args": [
      "sh"
    ],
    "env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ],
    "cwd": "/"
  },
  "root": {
    "path": "rootfs"
  },
  "hooks": {},
  "linux": {
    "resources": {
      "memory": {
        "limit": 0,
        "reservation": 0,
        "swap": 0
      },
      "cpu": {
        "shares": 0
      }
    }
  }
}
```

This is **wrong**. On Linux you **have** to include the `"namespaces": [ { "type": "mount" } ]` at the very least in order for it to even be possible for you to start inside a container. Preferably we should also add the PID namespace.

In addition, we have to add the default set of mounts mandated by opencontainers/runtime-spec#164. Namely we have to add `/proc`, `/sys` and `/dev` (runC will handle everything else for us).

cyphar:
The previous config generation produced a config which would cause runC
to run the container using the host filesystem (with no namespaces set
up). While we should try to be as agnostic as possible, we don't want to
generate config.json payloads that are known to be bad (and useless).

Fixes #76 

Signed-off-by: Aleksa Sarai <asarai@suse.com>

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -73,12 +73,92 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 		return nil, fmt.Errorf("%s: unsupported OS", c.OS)
 	}
 
-	var s specs.Spec
-	s.Version = specs.Version
-	// we should at least apply the default spec, otherwise this is totally useless
-	s.Process.Terminal = true
+	// TODO: Get this default from somewhere, rather than hardcoding it here.
+	// runC has a default as well, maybe the two should be consolidated in the
+	// runtime-spec (or in the runtime-tooling)?
+	s := specs.Spec{
+		Version: specs.Version,
+		Platform: specs.Platform{
+			OS:   "unknown",
+			Arch: "unknown",
+		},
+		Process: specs.Process{
+			Terminal: true,
+			Cwd:      "/",
+			Env:      []string{},
+			Args:     []string{},
+			User: specs.User{
+				UID: 0,
+				GID: 0,
+			},
+		},
+		Root: specs.Root{
+			Path:     "rootfs",
+			Readonly: false,
+		},
+		// These are all required by the runtime-spec to be included. The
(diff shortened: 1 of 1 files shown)


image/config.go
```diff
@@ -73,12 +73,92 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 		return nil, fmt.Errorf("%s: unsupported OS", c.OS)
 	}
 
-	var s specs.Spec
-	s.Version = specs.Version
-	// we should at least apply the default spec, otherwise this is totally useless
-	s.Process.Terminal = true
+	// TODO: Get this default from somewhere, rather than hardcoding it here.
```
wking: I suggest you get the default [from a vendored `runtime-tools/generate`][1].

[1]: https://github.com/opencontainers/runtime-tools/blob/bb040483d65d8731a2589f5d20c8b2367323219d/generate/generate.go#L34

cyphar: I ended up implementing cyphar/umoci#13 so I'll probably end up just merging that here and dropping the old code.

coolljt0725:
I'm +1 with this

vbatts:
I'm +1 for this, but it looks like `make lint` is failing.

q384566678:
@cyphar I passed the test after I updated it. Can you update it?

PR #78, opened 2016-11-07
main.go:Modify some information

q384566678:

### DIFF
--- a/cmd/oci-create-runtime-bundle/main.go
+++ b/cmd/oci-create-runtime-bundle/main.go
@@ -133,7 +133,7 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 	}
 
 	if err != nil {
-		v.stderr.Printf("unpacking failed: %v", err)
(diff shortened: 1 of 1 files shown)


xiekeyang:
Previous description should be acceptable, because it want to unpack source file to the rootfs.
Refer to https://github.com/opencontainers/image-tools/pull/47#discussion_r84600849

q384566678:
updated,But this way, and [this note](https://github.com/opencontainers/image-tools/blame/master/cmd/oci-create-runtime-bundle/main.go#L41) will not contradict each other, in addition，I think this is need to modify the error message

xiekeyang:
It looks goo to me.

wking:
c8145385 looks fine to me, although I'm fine with the previous language too.

q384566678:
@stevvooe  @vbatts  @philips PTAL

coolljt0725:
LGTM

xiekeyang:
LGTM

PR #79, opened 2016-11-08
create-main.go: If the dest file does not exist to create.

q384566678:

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -180,6 +180,16 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
 		return err
 	}
 
+	if _, err = os.Stat(dest); err != nil {
+		if os.IsNotExist(err) {
+			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
+				return err2
+			}
+		} else {
+			return err
+		}
+	}
+
 	err = m.unpack(w, filepath.Join(dest, rootfs))
 	if err != nil {
 		return err
--- a/cmd/oci-create-runtime-bundle/main.go
+++ b/cmd/oci-create-runtime-bundle/main.go
@@ -109,11 +109,6 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if _, err := os.Stat(args[1]); os.IsNotExist(err) {
-		v.stderr.Printf("destination path %s does not exist", args[1])
-		os.Exit(1)
-	}
-
 	if v.typ == "" {
 		typ, err := image.Autodetect(args[0])
 		if err != nil {


cmd/oci-unpack/main.go
```diff
@@ -101,6 +101,11 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	if _, err := os.Stat(args[1]); os.IsNotExist(err) {
+		v.stderr.Printf("destination path %s does not exist", args[1])
+		os.Exit(1)
```
wking: Is this a problem?  Why don't we create the directory if it is missing?  Although I'd rather `os.Mkdir` instead of `os.MkdirAll` to avoid overreacting to typo'd paths.  This would give us the same semantics as `touch(1)` and a host of other tools.

wking: Yeah, but don't create it if you'd have to make more than one new directory.

xiekeyang: > Is this a problem? Why don't we create the directory if it is missing? 

Agree, and it is what current design has done. Please don't change it.

xiekeyang: > I think here, either create, or error, because the following also used
It is created already in next steps, don't create here.

wking: It looks like the current code goes with `MkdirAll` as part of the [layer][1] [unpacking][2].  I think we probably want to transition that to a `Mkdir` approach for the `dest` directory, but I agree that the error floated here in 7e059a6 is too strict.

[1]: https://github.com/opencontainers/image-tools/blob/27d5fd6616cc7cd4de36521e9a0c9c8cfc8234ca/image/manifest.go#L169
[2]: https://github.com/opencontainers/image-tools/blob/27d5fd6616cc7cd4de36521e9a0c9c8cfc8234ca/image/manifest.go#L201

q384566678:
@wking @vbatts @philips @xiekeyang @coolljt0725 PTAL

image/image.go
```diff
@@ -180,6 +180,16 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
 		return err
 	}
 
+	if _, err = os.Stat(dest); err != nil {
+		if os.IsNotExist(err) {
+			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
```
q384566678: updated

wking:
4ecd0d39 looks good to me.  I think we still need directory-creation
work inside the unpacker, but that can happen in a later PR.

q384566678:
@stevvooe @vbatts @xiekeyang @coolljt0725 PTAL

xiekeyang:
LGTM

q384566678:
@coolljt0725 @stevvooe @vbatts PTAL

image/image.go
```diff
@@ -180,6 +180,16 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
 		return err
 	}
 
+	if _, err = os.Stat(dest); err != nil {
+		if os.IsNotExist(err) {
+			if err2 := os.Mkdir(dest, 0755); err2 != nil {
```
stevvooe: The UX here is going to be annoying without using `MkdirAll`.

Either we drop this and have a good error message like "destination does not exist, create it" or we use `MkdirAll`. Using `Mkdir` here is just annoying.

q384566678:
updated

stevvooe:
LGTM

q384566678:
ping @xiekeyang @coolljt0725

xiekeyang:
LGTM

Issue #15, opened 2016-09-21
cmd/oci-image-tool: is having empty refs/ a valid image?


philips:
_From @runcom on September 9, 2016 8:40_

$SUBJECT and basically, is it valid to have stuff under `blobs/` not referenced? Can I still package it up and validation is ok given there are no refs/? Should we print a warning (which can be silenced)?

_Copied from original issue: opencontainers/image-spec#287_

philips:
_From @wking on September 9, 2016 9:4_

On Fri, Sep 09, 2016 at 01:40:15AM -0700, Antonio Murdaca wrote:
> $SUBJECT

I think so.  At least I don't see any “and there MUST be at least one
ref” language.  And if you cared about running images from the layour
that would be “and there MUST be at least one unpackable ref”
language.  And then you have to match the oci-layout version with the
media types that were unpackable when it was current (or is validity
in the eye of the validator?)… This is a bowl that I do not want to
fathom ;).

Maybe folks are just using an image-layout to ship some missing blobs.

I don't see any incentive to image-authors to publish ref-less blobs
and then pretend they are runnable, so I don't see a need to get into
the business of restricting refs.

> … and basically, is it valid to have stuff under `blobs/` not
> referenced?

Yes [1].

> Should we print a warning?

Warnings for “there are no refs” and “dangling blob sha256-abc…” sound
useful to me.  Maybe put the first behind ‘--log-level warn’ and the
second behind ‘--log-level info’ based on how likely I think they are
to be a mistake and how many lines will get printed if they exist ;).

[1]: https://github.com/opencontainers/image-spec/blame/e6a431f1584468f112a52c01a02d326199b99021/image-layout.md#L46

philips:
_From @vbatts on September 13, 2016 19:20_

Both seem valid to me

philips:
_From @runcom on September 13, 2016 19:25_

Alright, can we keep this open so we won't forget about adding some sort of warning/log when this happens

xiekeyang:
I think it could be closed by https://github.com/opencontainers/image-spec/pull/533

PR #83, opened 2016-11-14
image.go:change out.Print to fmt.Errorf

q384566678:
I think we should return an error here, if it is to use WARING, then the validate function will return nil, and ultimately make the verification results are wrong

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -57,7 +57,7 @@ func validate(w walker, refs []string, out *log.Logger) error {
 		// TODO(runcom): ugly, we'll need a better way and library
 		// to express log levels.
 		// see https://github.com/opencontainers/image-spec/issues/288
-		out.Print("WARNING: no descriptors found")
+		return fmt.Errorf("no descriptors found")
 	}
 
 	if len(refs) == 0 {


wking:
On Sun, Nov 13, 2016 at 10:35:56PM -0800, Zhou Hao wrote:
> I think we should return an error here…

Previous discussion in #287 (maybe moved to #15) and [1].  I don't see
spec language requiring an image-layout to contain refs, and I'd
rather avoid a requirement like that.  I'm fine with a flag to turn
warnings to errors in this case (--common-use-cases? ;), but I don't
think we want this to be part of formal compliance testing.

[1]: https://github.com/opencontainers/image-spec/pull/279#r78030315

q384566678:
> zhouhao@zz:~/image-tools$ ./oci-image-validate --type imageLayout ../busybox-b
WARNING: no descriptors found
../busybox-b:OK

If not change, we are in accordance with the above command to verify an empty folder, it will only display warning, but it will still be OK, this is wrong

wking:
On Mon, Nov 14, 2016 at 05:27:12PM -0800, Zhou Hao wrote:
> > zhouhao@zz:~/image-tools$ ./oci-image-validate --type imageLayout ../busybox-b
> WARNING: no descriptors found
> ../busybox-b:OK
>
> If not change, we are in accordance with the above command to verify
> an empty folder, it will only display warning, but it will still be
> OK, this is wrong

You can have warnings in a valid image (that's the difference between
warnings and errors).  GCC has -Werror to turn all warnings into
errors, and I'm fine with something similar in image-tools.  But
without a way to toggle the error-ness of these warnings, I don't
think we should be erroring out where there is no spec requirement
being violated.

xiekeyang:
I debug it. the Warning happens when oci layout lack `refs` descriptor.
And [layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md#oci-image-layout-specification) spec declare that 
> The image layout has two top level directories:
"blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
"refs" contains descriptors. Commonly pointing to an image manifest or an image manifest list.

So I think it should return error instead warning in this case.

LGTM

wking:
On Tue, Nov 15, 2016 at 07:08:13PM -0800, xiekeyang wrote:
> And layout spec declare that
>
> > The image layout has two top level directories:
> > "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
> > "refs" contains descriptors. Commonly pointing to an image manifest or an image manifest list.
> 
> So I think it should return error instead warning in this case.

Are you interpreting “contains descriptors” as “MUST contain at least
one descriptor”?  I don't read it that way.  If you *do* think there's
a case for that interpretation, it's probably worth making that change
via an image-spec PR, instead of adding strict image-tools tests based
on an ambiguous spec requirement.

Mashimiao:
@xiekeyang   If `LGTM` is not the beginning of  comment, it doesn't make any sense.  you need re-LGTM

xiekeyang:
@wking 
> Are you interpreting “contains descriptors” as “MUST contain at least
one descriptor”?  ...

I think we are discussing if `ref` is necessary for OCI image. IMO, I thinks it is necessary, as the identifier. Right?

wking:
On Tue, Nov 15, 2016 at 10:51:37PM -0800, xiekeyang wrote:
> I think we are discussing if `ref` is necessary for OCI image. IMO,
> I thinks it is necessary, as the identifier. Right?

I've filed opencontainers/image-spec#459 to add clarity to the
image-layout requirements.  If that looks good to you, maybe our
difference is what the image-layout is getting validated as.  I've
been resisting errors here because I was thinking we were validating
it *as an image-layout*.  Your “OCI image” comment suggests you are
validating it as an *OCI Image* [1].  I agree that you need a ref,
manifest, and config, and (currently) at least one layer to be a valid
OCI Image.

When you give validation an explicit list of refs, it's fairly clear
you want those validated as OCI Images.

When you just give it an image-layout, it's not clear (to me) that you
are requiring it to contain an OCI Image.  I'd expect you to validate
the image-layout, and then validate any refs that point to manifests
or manifest lists as OCI Images.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc2/README.md#L25

xiekeyang:
As most folks accept opencontainers/image-spec#459, this PR should be closed.

wking:
On Wed, Nov 16, 2016 at 06:32:32PM -0800, xiekeyang wrote:
> As most folks accept opencontainers/image-spec#459, this PR should
> be closed.

I think there may still be room for improved validation here if we can
figure out a UI that easily distinguishes between validating an
image-layout and requiring an OCI Image to be contained within that
image-layout [1].  --has-at-least-one-oci-image is a bit heavy handed,
but maybe something along those lines would cover the use-case you
were getting at here?

[1]: https://github.com/opencontainers/image-tools/pull/83#issuecomment-260991412

PR #50, opened 2016-10-25
Validate oci-layout version

coolljt0725:
OCI image specification has required that
`"oci-layout" MUST contain a JSON object with a version field {"imageLayoutVersion": "1.0.0"}`,
the image tool should follow this rule.

Signed-off-by: Lei Jitang leijitang@huawei.com

PR #84, opened 2016-11-15
image: add image layout validation 

q384566678:
Increase the validation of the files contained in the layout.
Including `blobs` , `oci-layout` and `index.json` files.

### DIFF
--- a/image/layout.go
+++ b/image/layout.go
@@ -1,0 +1,104 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+func layoutValidate(w walker) error {
+	var blobsExist, indexExist, layoutExist bool
+
+	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
+		if strings.EqualFold(path, "blobs") {
+			blobsExist = true
+			if !info.IsDir() {
+				return fmt.Errorf("blobs is not a directory")
+			}
+
+			return nil
+		}
+
+		if strings.EqualFold(path, "index.json") {
+			indexExist = true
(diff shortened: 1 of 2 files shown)


q384566678:
@wking @xiekeyang @stevvooe @philips @coolljt0725 @vbatts PTAL

xiekeyang:
`Autodetect` is just distinguishing between types. In next steps(unpack or validation) objects will be found and determined. That can make process complete.
So this modification seems to be superfluous.

coolljt0725:
https://github.com/opencontainers/image-spec/pull/459  has clarify that `blobs` and `refs` MUST exist, but for the `oci-layout`, the file name is not clear if it must be `oci-layout` at the moment.

wking:
I think the presence of `oci-layout` with a recognized version is enough to decide the target is attempting to be an image-layout.  The presence of `blobs` and `refs` directories are part of determining whether that image-layout is valid/usable, but they probably don't need to be part of autodetection.

image/autodetect.go
```diff
@@ -43,6 +50,11 @@ func Autodetect(path string) (string, error) {
 	}
 
 	if fi.IsDir() {
+		for _, file := range Layoutfile {
+			if _, err = os.Stat(path + file); os.IsNotExist(err) {
+				return "", errors.New("unknown media type")
```
stevvooe: Make this more specific.

Mashimiao:
If we just support image layout, image, manifest, manifestlist and image config, there is no need to add layout validation in autodetect. Because if the input is a directory, it should be considered as a image layout.  Only image layout could be a directory in supported type validation
I think move the code in this PR into `image.ValidateLayout()` will be better.

image/autodetect.go
```diff
@@ -34,6 +36,13 @@ const (
 	TypeConfig       = "config"
 )
 
+// Layoutfile realize the image files that are included in the image layout type
+var Layoutfile = []string{
+	"blobs",
+	"oci-layout",
+	"refs",
```
vbatts: this is changed now. There is no longer a ./refs/ directory. And there is an `index.json` file. See the updated image-layout.md doc.

Mashimiao:
@q384566678  The Travis CI failed, I think you need to fix TestValidateLayout in image_test.go

q384566678:
> The Travis CI failed, I think you need to fix TestValidateLayout in image_test.go

I'm doing, but there are some content to wait for `image-spec` certain.

q384566678:
Some methods in #50  are used to update the validation of layout. 
@xiekeyang @coolljt0725 @cyphar @stevvooe @vbatts  PTAL

q384566678:
reping @opencontainers/image-tools-maintainers  I hope you can take a moment to review this PR, which is an important validation that is missing.

image/layout.go
```diff
@@ -0,0 +1,104 @@
...
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+var blobsExist, indexExist, layoutExist bool
```
coolljt0725: can we put these into `layoutValidate`?

q384566678: ok, updated.

coolljt0725:
LGTM   tested, it works for me

xiekeyang:
LGTM

PR #85, opened 2016-11-17
config.go:Drop local rootfs.type validation

q384566678:
In this type of authentication is not required, because the json file has been identified [type value](https://github.com/opencontainers/image-spec/blame/master/schema/defs-config.json#L81) ，if not the case, it will never perform type validation.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -53,10 +53,7 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 		if err := json.Unmarshal(buf, &c); err != nil {
 			return err
 		}
-		// check if the rootfs type is 'layers'
-		if c.RootFS.Type != "layers" {
-			return fmt.Errorf("%q is an unknown rootfs type, MUST be 'layers'", c.RootFS.Type)
-		}
+
 		return errEOW
 	}); err {
 	case nil:


wking:
On Thu, Nov 17, 2016 at 12:09:44AM -0800, Zhou Hao wrote:
>   * config.go:Changes the verifying position of layers

This is checked in the JSON Schema since
opencontainers/image-spec#358, so I'd rather just drop the local check
entirely.  The current, generic JSON-Schema-failure:

  errors.Wrapf(err, "%s: config validation failed", path)

will give you a message about the invalid rootfs.type without us
needing special-case JSON-Schema-failure handling for this situation.

q384566678:
> This is checked in the JSON Schema since opencontainers/image-spec#358, so I'd rather just drop the local check entirely. The current, generic JSON-Schema-failure:

I think if the type validation fails or need a separate error message, because in accordance with the [provisions of the meaning](https://github.com/opencontainers/image-spec/blame/master/config.md#L150), the error message should be more detailed.

wking:
On Thu, Nov 17, 2016 at 05:38:49PM -0800, Zhou Hao wrote:
> I think if the type validation fails or need a separate error
> message, because in accordance with the [provisions of the
> meaning](https://github.com/opencontainers/image-spec/blame/master/config.md#L150),
> the error message should be more detailed.

The JSON Schema error from the current master looks like:

  $ git checkout origin/master
  $ git describe --always
  5702049
  $ make tools
  $ cat >config.json <<EOF
  {
    "rootfs": {
      "type": "something-else",
      "diff_ids": []
    }
  }
  EOF
  $ ./oci-image-validate --type config config.json
  config.json: validation failed: rootfs.type: rootfs.type must be one of the following: "layers"

That's pretty specific.  It doesn't link to the relevant spec line,
but then neither does your patch.  And the JSON Schema message is
invoked regardless of how we find the config, while the line you're
shifting here only fires when you're fetching the config from a walker
(my oci-image-validate line above is getting the file directly,
without going through a walker).

q384566678:
> $ ./oci-image-validate --type config config.json 
config.json: validation failed: rootfs.type: rootfs.type must be one of the following: "layers"

So to say, this type of authentication is not necessary

wking:
The diff for 3aaa0b8 looks good to me, although the commit message
should probably be updated to something like:

  config.go: Drop local rootfs.type validation

  This is handled in image-spec since opencontainers/image-spec#358
  (v1.0.0-rc2).

coolljt0725:
LGTM we could drop local rootfs.type validation since it has been done in image-spec schema

xiekeyang:
LGTM

PR #87, opened 2016-11-18
Disallow self-LGTMs

coolljt0725:
according https://github.com/opencontainers/project-template/pull/13
we should disallow self-LGTMs


stevvooe:
LGTM

xiekeyang:
LGTM

PR #89, opened 2016-11-28
Modify the ref information

q384566678:
According to some [norms](https://github.com/opencontainers/image-spec/blame/master/image-layout.md#L74) can be seen, ref can also point to the manifest list, here only to modify some of the error message, some code error will continue to update in #51 .

I think it's more appropriate to change ref to tag. 

### DIFF
--- a/man/oci-image-tool-create.1.md
+++ b/man/oci-image-tool-create.1.md
@@ -18,7 +18,7 @@ runtime-spec-compatible `dest/config.json`.
   Print usage statement
 
 **--ref**=""
-  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+  The ref pointing to the manifest of the OCI image. This must be match 'org.opencontainers.image.ref.name' annotations in index.json. (default "v1.0")
 
 **--rootfs**=""
   A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -90,7 +90,7 @@ var createCommand = cli.Command{
 		cli.StringFlag{
 			Name:  "ref",
 			Value: "v1.0",
-			Usage: "The ref pointing to the manifest of the OCI image. This must be present in the 'refs' subdirectory of the image.",
+			Usage: "The ref pointing to the manifest of the OCI image. This must be match 'org.opencontainers.image.ref.name' annotations in index.json.",
 		},
 		cli.StringFlag{
 			Name:  "rootfs",
--- a/cmd/oci-image-tool/unpack.go
+++ b/cmd/oci-image-tool/unpack.go
@@ -86,7 +86,7 @@ var unpackCommand = cli.Command{
 		cli.StringFlag{
 			Name:  "ref",
 			Value: "v1.0",
-			Usage: "The ref pointing to the manifest of the OCI image. This must be present in the 'refs' subdirectory of the image.",
+			Usage: "The ref pointing to the manifest of the OCI image. This must be match 'org.opencontainers.image.ref.name' annotations in index.json.",
 		},
 		cli.StringFlag{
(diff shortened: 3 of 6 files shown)


image/image.go
```diff
@@ -120,7 +120,7 @@ func unpack(w walker, dest, refName string) error {
 		return err
 	}
 
-	if err = ref.validate(w, validRefMediaTypes); err != nil {
+	if err = ref.validate(w, []string{v1.MediaTypeImageManifest}); err != nil {
```
coolljt0725: the ref could refer to `manifest` or  `manifest list`, I don't think we need this change.

q384566678: But in this place must be `manifest`

coolljt0725: I don't think so. If the `ref` of an OCI image  refer to a manifest list, and the the manifest list refer to a manifest,  the `findManifest` below are supposed to get the `manifest` from the `manifest list`. With your changes, this process are break.

q384566678: `manifest list` points to only a few special `manifests`, but specify its `platform`, there is no other property requirements. I do not think this is what you want to get, and in the instructions, `ref` can only point to the `manifest`.

q384566678: In the pr #51 ,Only in the validation function need to confirm the manifest and manifestlist, and I just change the unpack and create-runtime-bundel function

coolljt0725: > manifest list points to only a few special manifests, just specify its platform, there is no other property 

I don't know if you understand the `manifest list` correctly :).  How could you unpack a OCI image if the `ref`( for more information about ref https://github.com/opencontainers/image-spec/blob/master/image-layout.md) points to  manifest list ? and how do you know the `ref` is always point to a manifest ?

q384566678: You are right, I was misled by some wrong information, and now I have this information modified.

coolljt0725:
@q384566678  image-tool can't handle manifest list at the moment, let's waiting the image-tool for adding this function and then let's has these changes  in this pr, WDYT?

q384566678:
> let's waiting the image-tool for adding this function and then let's has these changes in this pr, WDYT?

I increase the corresponding function in the pr #51 , because of some rely on pr #51 function, so can't update in the pr.

q384566678:
@jonboulle @stevvooe @vbatts @philips  PTAL

q384566678:

q384566678:
@xiekeyang @coolljt0725 @vbatts @stevvooe PTAL

coolljt0725:
let's make the image-tool can handle manifest list first then update the docs.  close at the moment and will reopen once the image-tool can handle manifest list

q384566678:
Since the verification of the index has been improved in #51 , so I reopen this.

man/oci-image-tool-create.1.md
```diff
@@ -18,7 +18,7 @@ runtime-spec-compatible `dest/config.json`.
   Print usage statement
 
 **--ref**=""
-  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+  The ref pointing to the manifest or the image index of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
```
q384566678: updated. And I think it's more appropriate to change ref to tag.

man/oci-image-tool-create.1.md
```diff
@@ -17,8 +17,8 @@ runtime-spec-compatible `dest/config.json`.
 **--help**
   Print usage statement
 
-**--ref**=""
-  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+**--tag**=""
+  The tag pointing to the manifest or the image index of the OCI image. his must be matching 'org.opencontainers.image.ref.name' annotations in index.json. (default "v1.0")
```
Mashimiao: The spec always uses ref, and I prefer to ref.  And ref must be pointing to a manifest, will not be an image index, we can get the manifest var image index. So I think the change is not correct.

q384566678:
Because #169  has been merged, so close this pr.

PR #90, opened 2016-11-30
manifest.go:remove the size of the layers of judgment

q384566678:
According to [this](https://github.com/opencontainers/image-spec/blame/master/schema/image-manifest-schema.json#L26) ，we can see the minimum length of the layer has been verified in JSON, so the code is redundant.

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -59,10 +59,6 @@ func findManifest(w walker, d *descriptor) (*manifest, error) {
 			return err
 		}
 
-		if len(m.Layers) == 0 {
-			return fmt.Errorf("%s: no layers found", path)
-		}
-
 		return errEOW
 	}); err {
 	case nil:


wking:
On Tue, Nov 29, 2016 at 05:53:42PM -0800, Zhou Hao wrote:
> … the minimum length of the layer has been verified in JSON, so the
> code is redundant.

That landed in opencontainers/image-spec#466, which is currently
unreleased.  Once image-spec cuts rc3 (opencontainers/image-spec#477)
and image-tools bumps to pick that up, f5d7ced3 will look good to me.

q384566678:
@coolljt0725 @xiekeyang PTAL

q384566678:

coolljt0725:
@q384566678  we need to bump the image-spec to rc3 first

q384566678:
> we need to bump the image-spec to rc3 first

@coolljt0725 I don't quite understand why .

coolljt0725:
@q384566678  as said by @wking  in https://github.com/opencontainers/image-tools/pull/90#issuecomment-264108575

q384566678:
@coolljt0725 That PR(opencontainers/image-spec#477) has been merged, it has become rc3.

coolljt0725:
@q384566678  but image-tool doesn't bump it yet. https://github.com/opencontainers/image-tools/blob/master/glide.yaml#L3

q384566678:
@coolljt0725 It can be directly modified like this?

q384566678:
@coolljt0725 @xiekeyang @vbatts @stevvooe rebased #102 PTAL

q384566678:
cc @coolljt0725 @xiekeyang

coolljt0725:
LGTM

xiekeyang:
LGTM

PR #93, opened 2016-12-02
manifest.go: Remove the extra judgment for layers

q384566678:
In the process of the validation have been carried out for the validation of the [layers,](https://github.com/opencontainers/image-tools/blame/master/image/manifest.go#L83) So they are not needed in the unpack function.


q384566678:

q384566678:
@stevvooe @xiekeyang @coolljt0725 @vbatts  PTAL

stevvooe:
LGTM

coolljt0725:
LGTM

PR #95, opened 2016-12-06
Fix information

q384566678:

### DIFF
--- a/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
+++ b/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
@@ -6,7 +6,7 @@ oci-create-runtime-bundle \- Create an OCI runtime bundle
 
 # SYNOPSIS
 **oci-create-runtime-bundle** [src] [dest] [flags]
-**oci-create-runtime-bundle** [--help|--version]
+**oci-create-runtime-bundle** [--help|-v|--version]
 
 # DESCRIPTION
 `oci-create-runtime-bundle` validates an application/vnd.oci.image.manifest.v1+json and unpacks its layered filesystem to `dest/rootfs`, although the target directory is configurable with `--rootfs`. See **oci-unpack**(1) for more details on this process.
@@ -18,16 +18,16 @@ runtime-spec-compatible `dest/config.json`.
 **--help**
   Print usage statement
 
-**--ref**
+**--ref**=""
   The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
 
-**--rootfs**
+**--rootfs**=""
   A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
 
-**--type**
+**--type**=""
   Type of the file to unpack. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
 
-**--version**
+**-v**, **--version**
   Print version information and exit.
 
 # EXAMPLES
--- a/cmd/oci-image-validate/oci-image-validate.1.md
+++ b/cmd/oci-image-validate/oci-image-validate.1.md
@@ -6,7 +6,7 @@ oci-image-validate \- Validate one or more image files
 
 # SYNOPSIS
 **oci-image-validate** FILE... [flags]
-**oci-image-validate** [--help|--version]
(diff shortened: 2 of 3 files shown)


q384566678:

cmd/oci-image-validate/oci-image-validate.1.md
```diff
@@ -16,18 +16,18 @@ oci-image-validate \- Validate one or more image files
 **--help**
   Print usage statement
 
-**--ref** NAME
+**--ref**=[] NAME
```
Mashimiao: This is not a valid format. `**--ref**=[NAME...]` or `**--ref**=[]` will be OK

q384566678:
cc @xiekeyang @coolljt0725

xiekeyang:
LGTM

cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
```diff
@@ -18,17 +18,17 @@ runtime-spec-compatible `dest/config.json`.
...
   A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
 
-**--type**
+**--type**=""
   Type of the file to unpack. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
 
-**--version**
-  Print version information and exit.
+**-v**, **--version**=*true*|*false*
+  Print version information and exit. Default is false.
```
coolljt0725: I think `=*true*|*false*` and `Default is false` is not necessary.

cmd/oci-image-validate/oci-image-validate.1.md
```diff
@@ -16,18 +16,18 @@ oci-image-validate \- Validate one or more image files
...
   Defaults to `v1.0`.
   Only applicable if type is image or imageLayout.
 
-**--type**
+**--type**=""
   Type of the file to validate. If unset, oci-image-validate will try to auto-detect the type. One of "imageLayout,image,manifest,manifestList,config"
 
-**--version**
-  Print version information and exit.
+**-v**, **--version**=*true*|*false*
```
coolljt0725: same suggestion as above

cmd/oci-unpack/oci-unpack.1.md
```diff
@@ -15,14 +15,14 @@ oci-unpack \- Unpack an image or image source layout
...
+**--ref**=""
   The ref pointing to the manifest to be unpacked. This must be present in the "refs" subdirectory of the image. (default "v1.0")
 
-**--type**
+**--type**=""
   Type of the file to unpack. If unset, oci-unpack will try to auto-detect the type. One of "imageLayout,image"
 
-**--version**
-  Print version information and exit.
+**-v**, **--version**=*true*|*false*
```
coolljt0725: also same here

cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
q384566678: This is the optional parameters of the two values, I think there can be here. This allows the user to better understand this parameter.

coolljt0725: I don't think user will use `oci-create-runtime-bundle -v=false`, it's a wrong use case and it will print the help messages.
```
$ oci-create-runtime-bundle -v=false
both src and dest must be provided
Usage:
  oci-create-runtime-bundle [src] [dest] [flags]

Flags:
      --ref string      The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
      --rootfs string   A directory representing the root filesystem of the container in the OCI runtime bundle.
It is strongly recommended to keep the default value. (default "rootfs")
      --type string     Type of the file to unpack. If unset, oci-create-runtime-bundle will try to auto-detect the type. One of "imageLayout,image"
  -v, --version         Print version information and exit

```

and juse check some other common linux tools like `man fdisk`
```
-V, --version
              Display version information and exit.
```

and also the runtime tools https://github.com/opencontainers/runtime-tools/blob/master/man/oci-runtime-tool.1.md

q384566678: OK，updated

coolljt0725:
LGTM

xiekeyang:
LGTM

PR #97, opened 2016-12-09
Returns an error if the condition is not met

q384566678:
When some of the conditions do not meet, you can not only return nil, but also return an error message to terminate the program.

### DIFF
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -54,7 +54,7 @@ func listReferences(w walker) (map[string]*descriptor, error) {
 
 	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() || !strings.HasPrefix(path, "refs") {
-			return nil
+			return errors.New("no descriptors found")
 		}
 
 		var d descriptor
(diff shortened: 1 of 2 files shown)


wking:
On Thu, Dec 08, 2016 at 10:34:33PM -0800, Zhou Hao wrote:
> When some of the conditions do not meet, you can not only return
> nil, but also return an error message to terminate the program.

This PR is very similar to #83, and I recommend closing this as well
(although there's still room for improving the UI [1]).  Having no
references is *not* an error.  It is surprising, hence the warning,
but an image without refs is technically legal and has some legitimate
use cases.  More background in opencontainers/image-spec#459,
opencontainers/image-spec#287, and the #83 discussion.

On the other hand, finding a subdirectory under `refs/` *may* be an
error.  The current spec wording has a few instances of “the refs
subdirectories”, including the charset restriction [2].  But my guess
is that the intention was to restrict refs to a single directory.  I'm
not sure if the intention for refs/somedir/ was to ignore it or error
out though.  An image-spec PR to clarify this point is probably in
order.

[1]: https://github.com/opencontainers/image-tools/pull/83#issuecomment-261157131
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/image-layout.md

q384566678:
Well, it really should not be so modified here, but I always feel that this place to determine some problems.

PR #98, opened 2016-12-14
Increase the authentication type

q384566678:
When I use the validate function, and verify if it is through the automatic judgment function judgment type, do not know what is the type of validation, I think in this authentication type prompt better.

### DIFF
--- a/cmd/oci-image-tool/validate.go
+++ b/cmd/oci-image-tool/validate.go
@@ -93,6 +93,8 @@ func validatePath(name string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


q384566678:

Mashimiao:
it's better to clear tell user the type belongs to which file

q384566678:
@vbatts @xiekeyang @stevvooe @coolljt0725 PTAL

stevvooe:
Why are still putting so much work into autodetection?

q384566678:
> Why are still putting so much work into autodetection?
@stevvooe Here is only to increase the type of authentication information. So, what is the problem?

PR #99, opened 2016-12-15
validate.md: delete the ref's default value

q384566678:
There is no default value given to the ref option, so should remove this statement.


q384566678:

coolljt0725:
LGTM

stevvooe:
LGTM

PR #101, opened 2016-12-22
Increase the generate function

q384566678:
Increase the generate function，you can create an OCI-compliant imageLayout based on the address you enter.Now just create the imageLayout, after the completion of the code can also create other types (I do not know if these are not necessary).


q384566678:
This is my initial idea, I will continue to improve the PR, such as documentation, other types of creation, etc.

image/image.go
```diff
@@ -198,3 +200,44 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
...
+	}()
+
+	il := imageLayout{
+		rootDir:  dest,
+		layout:   layoutStr,
+		ref:      refTag,
+		manifest: manifestStr,
+		config:   configStr,
+		tarList: []tarContent{
+			tarContent{&tar.Header{Name: "example", Size: 7, Mode: 0600}, []byte("example")},
```
wking: I don't think we want to hard-code any tar content for this command.

Also, I think we want to give users a way to pick different image-layout backends (e.g. tar-backed image-layouts, directory-backed image-layouts, …).  For example, the `oci-image-init` I've floated in #5 [takes a `--type` option][1].

[1]: https://github.com/opencontainers/image-tools/pull/5/files#diff-0cf1d5acd82a85146cf1eaf17df69bbcR21

q384566678:
@stevvooe @coolljt0725 @xiekeyang @vbatts PTAL

q384566678:
ping @xiekeyang @coolljt0725 @stevvooe @vbatts @jonboulle PTAL ,thanks.

Makefile
```diff
@@ -7,8 +7,8 @@ EPOCH_TEST_COMMIT ?= v0.2.0
 TOOLS := \
 	oci-create-runtime-bundle \
 	oci-image-validate \
-	oci-unpack
-MAN := $(TOOLS:%=%.1)
```
wking: You probably don't want to remove this line ;).

.gitignore
```diff
@@ -1,3 +1,4 @@
 /oci-create-runtime-bundle
 /oci-unpack
 /oci-image-validate
+/oci-generate
```
wking: The current position of `oci-unpack` is unfortunate, but I think we should alphabetize these entries (and the entries in the Makefile's `TOOLS`).

image/image.go
```diff
@@ -208,3 +209,51 @@ func createRuntimeBundle(w walker, dest, refName, rootfs string) error {
...
+func Generate(dest string) error {
+	err := GenerateLayout(dest)
+	if err != nil {
+		return err
+	}
+
+	return Compress(dest)
+}
+
+//GenerateLayout create an imageLayou that conforms to the OCI specification on a given path
```
wking: “Layou” → “Layout”.  And do you have any plans to provide a way to generate a tar-backed image layout?  I [think we want to support that][1].

[1]: https://github.com/opencontainers/image-tools/pull/101#discussion_r93663062

image/generate.go
```diff
@@ -0,0 +1,362 @@
...
+
+	return descriptor{
+		Digest: fmt.Sprintf("%x", hash.Sum(nil)),
+		Size:   size,
+	}, nil
+}
+
+//Compress use gzip to compress tar.gz
+func Compress(src string) error {
+	d, err := os.Create(src + ".tar")
```
wking: A function that compresses an on-disk directory into an on-disk file (I don't actually see any code in here about compressing with gzip) doesn't seem like the friendliest public API.  And you don't seem to be using this function anywhere.  So I suspect you should drop `Compress` and `compress` from this PR.

image/generate.go
```diff
@@ -0,0 +1,362 @@
...
+	desc, err := createHashedBlob(name)
+	if err != nil {
+		return descriptor{}, err
+	}
+
+	desc.MediaType = v1.MediaTypeImageLayer
+	return desc, nil
+}
+
+func createTarBlob(name string, list []tarContent) error {
```
wking: I'm holding out hope that #8 or something like that lands.  Until we get a public API for creating layers, I'd rather not drag `createTarBlob` out of `image_test.go`.

image/generate.go
```diff
@@ -0,0 +1,362 @@
...
+		return err
+	}
+
+	// create image layout file
+	err = createLayoutFile(il.rootDir)
+	if err != nil {
+		return err
+	}
+
+	// create image layer blob file.
```
wking: I'd prefer this function be public, [create whatever boilerplate the spec requires image layouts to have, and then stop][1].  Filling in example layers, JSON blobs, and refs isn't useful (we can always distribute the example image directly if that content is useful).  What I think users want (among other things) is a tool to create a new, *empty* image layout so they can start populating it with their content.  And that's what #5's [`oci-image-init image-layout …`][2], [`CreateDir`][3], and [`CreateTarFile`][4] are doing.

[1]: https://github.com/opencontainers/image-tools/pull/101#discussion_r93663062
[2]: https://github.com/opencontainers/image-tools/pull/5/files#diff-0cf1d5acd82a85146cf1eaf17df69bbcR10
[3]: https://github.com/opencontainers/image-tools/pull/5/files#diff-1d84590cedd145294b2105cc459ef5f3R43
[4]: https://github.com/opencontainers/image-tools/pull/5/files#diff-745d1b74657ed8e87ac8c4e3f520f73aR212

PR #102, opened 2016-12-29
Delete does not exist properties

q384566678:
Make the following changes according to https://github.com/opencontainers/image-spec/pull/495


coolljt0725:
We need to bump the image-spec first

q384566678:
@coolljt0725 updated.Thank you so much for your suggestion.
@xiekeyang @stevvooe PTAL

coolljt0725:
@q384566678  AFAIK, these being removed field does in `v1.0.0-rc3`,   I think we should wait the https://github.com/opencontainers/image-spec/pull/495 being land on next release and then bump the image-spec.

stevvooe:
LGTM

xiekeyang:
LGTM

PR #103, opened 2017-01-06
Merging of subcommands

### LINKED ISSUE #96: About the merging of subcommands
I do not understand why the image-tools in the three sub-command split into three software, I think these three commands should be combined into a tool, after all, they belong to image-tools, I think the merger will Easy to understand and application, if I was a user, I would think that the use of a software implementation of the three commands than the use of three software to implement a command to be more convenient.
I currently have a plan to merge, don't know how you think .

q384566678:
This is some idea about #96 

Fixes #96 


vbatts:
i think conceptually this SGTM

xiekeyang:
SGTM. need rebase.

q384566678:
@vbatts @xiekeyang I think this change will be better. PTAL

xiekeyang:
@q384566678 It seems that `log-level` is in progress on #46 . In that patch, @stevvooe suggest that we should not use `log-level`, instead to use debug mode. And you seems not replace fmt.Print by `logrus` with `context`. So I think you seems to had better to remove this stuff from this patch.

man/oci-image-tool.1.md
```diff
@@ -0,0 +1,43 @@
...
+
+**--log-level**=LEVEL
+  Log level (panic, fatal, error, warn, info, or debug) (default: "error").
+
+**-v**, **--version**
+  Print version information.
+
+# COMMANDS
+**validate**
+  Validate one or more image files
```
stevvooe: What is an "image file"?

man/oci-image-tool.1.md
```diff
@@ -0,0 +1,43 @@
...
+  See **oci-image-tool-validate**(1) for full documentation on the **validate** command.
+
+**unpack**
+  Unpack an image or image source layout
```
stevvooe: What is an "image" or "image source layout"?

man/oci-image-tool.1.md
```diff
@@ -0,0 +1,43 @@
...
+  See **oci-image-tool-unpack**(1) for full documentation on the **unpack** command.
+
+**bundleCreate**
```
stevvooe: Let's avoid camelCase in our command designs. This should employ `<object> <verb>` or use dashes `bundle-create`.

cmd/oci-image-tool/unpack.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		fmt.Printf("unpacking failed: %v\n", err)
+	}
+	return err
+}
+
+var unpackCommand = cli.Command{
+	Name:   "unpack",
+	Usage:  "Unpack an image or image source layout",
+	Action: unpackHandle,
```
stevvooe: `unpackHandler`.

man/oci-image-tool.1.md
```diff
@@ -0,0 +1,43 @@
...
+
+**-v**, **--version**
+  Print version information.
+
+# COMMANDS
+**validate**
+  Validate the given file(s) against the OCI image specification
+  See **oci-image-tool-validate**(1) for full documentation on the **validate** command.
+
+**unpack**
```
stevvooe: This language still doesn't make the different between "unpack" and "create" clear.

stevvooe:
LGTM

Please follow up on the issues with `create` / `unpack`.

vbatts:
LGTM

xiekeyang:
@q384566678 @vbatts @stevvooe This PR call `logrus` function, without introduce it to `vendor`. It  leads build fail on `make tool`:
```
$ make tool 
go build -o oci-image-tool ./cmd/oci-image-tool
cmd/oci-image-tool/main.go:21:2: cannot find package "github.com/Sirupsen/logrus" in any of:
```
At least we should import `logrus` to `glide.yaml` and then `make update-deps`, the package can be generated in `vendor`.
Please take a review, is it Right?

PR #106, opened 2017-01-20
manpage: modify OPTIONS to FLAGS

Mashimiao:


xiekeyang:
I think change from `flag` to `option` is OK, but this should be accordant to tools usage.

Mashimiao:
I think the change does not make manpage to be not accordant with tools usage.

xiekeyang:
I mean help message of tools had better to be changed accordingly(`flag` to `option`), to be like
```
$ ./oci-image-validate --help
Usage:  oci-image-validate FILE... [options...]
Options: ...
```

Mashimiao:
Though I want to change all of `flags` to `options` which I think is more common.
But as cobra use flags instead of options as default, it seems better to change `OPTIONS` to `FLAGS` in manpage.
@xiekeyang  PTAL

xiekeyang:
LGTM

vbatts:
LGTM

PR #109, opened 2017-02-07
Increase the success tips execute the command

q384566678:

### DIFF
--- a/cmd/oci-create-runtime-bundle/main.go
+++ b/cmd/oci-create-runtime-bundle/main.go
@@ -146,5 +146,6 @@ func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	fmt.Println("creating succeeded")
 	os.Exit(0)
 }
--- a/cmd/oci-unpack/main.go
+++ b/cmd/oci-unpack/main.go
@@ -138,5 +138,6 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	fmt.Println("unpacking succeeded")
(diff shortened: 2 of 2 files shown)


cmd/oci-image-validate/main.go
```diff
@@ -145,6 +145,10 @@ func (v *validateCmd) Run(cmd *cobra.Command, args []string) {
 		exitcode = 1
 	}
 
+	if exitcode == 0 {
+		v.stdout.Println("validation succeeded")
```
xiekeyang: Preline has printed the OK message, this line seems redundant. Other changes of this PR seem good.

q384566678: OK is the determination of each arg, rather than the overall validate decision.
I would like to show that the entire order of judgment (args), rather than a single decision (arg).

runcom:
is this really needed :confused: exiting 0 seems nicer to me

q384566678:
> is this really needed :confused: exiting 0 seems nicer to me

@runcom I think it is necessary, because the user will not see `exiting 0`, increase the prompt statement, which allows us to more intuitive understanding of the operating state.

cyphar:
NACK. This is unnecessary.

q384566678:
Since everyone thinks this is not a good modification (although I do not think so), I will close this PR.

cyphar:
@q384566678 For some explanation of why outputting whether the return value is zero: In general, Unix tools don't work this way. An error message is usually output to `stderr`, but that error message should have some _useful_ information. "There was an error" is not useful information, because that's already contained in the return code.

A user can see the failure if they check `$?` or they have `set -e` (or hell, if they have a modern shell prompt which tells you whether the last command failed).

PR #108, opened 2017-02-07
manifest.go: add the layers validation type

q384566678:
Layer has several  [mediatype](https://github.com/opencontainers/image-spec/blame/master/specs-go/v1/mediatype.go#L28-L41) types, it should be completed all the mediatype.

Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

Issue #66, opened 2016-10-31
Validation strictness


wking:
Refs can point to whichever type they like (although they [usually point at manifests or manifest lists][1]).  How should we handle validation when a ref points at a single layer?  Or at a config?  Or at an image/png?  I see a few interesting settings:

* `--unpackable`, for “can image-tools unpack this content?”.  This would be true for manifests, manifest lists, and (once we land support for it) layers, but would not be true for configs (because diffIDs are not neccessarily fetchable from CAS).

* `--strict`, for “does the validated object only contain content which the OCI defines (or references) in the spec?”.  In that case, refs pointing to manifests, manifest lists, configs, or layers would all
  be legal.  Other types would trigger validation errors.

* `--base`, for “does the validated object only contain content which the OCI *requires* unpackers to support?”.  This would be true for manifests and layers, but not for manifest lists, configs, or other
  types.  There has been talk about [requiring unpacker support for manifest lists][2], we currently only [SHOULD support for them][3].  If we MUST support for them, then there may not be any need for `--base` (since it would cover the same types as `--strict`).

In cases where the user specified none of `--unpackable`, `--strict`, or `--base`, I'd rather warn (but not error) on this sort of thing.  And using TAP (as described in #60) gives us a convenient way to warn
without erroring.

[1]: https://github.com/opencontainers/image-spec/blame/5dbe4a342b6de21149f5c9f23c13b62ec1503082/image-layout.md#L15
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-19-21.00.log.html#l-74
[3]: https://github.com/opencontainers/image-spec/blame/5dbe4a342b6de21149f5c9f23c13b62ec1503082/manifest-list.md#L4

xiekeyang:
@wking 
If manifest, manifest list and single layer are unpackable? I'm not sure, but could you present some use case for that?

wking:
A use case for unpacking a single layer?  Maybe unpacking a manifest is not giving you the results you expect, and you want to step through the layers one by one.

PR #111, opened 2017-02-08
image: make layer validation work

cyphar:
Most recently, the validation was broken because the image-spec was
re-vendored but the code handling MediaType validation was not updated.

This change (among others) broke umoci's tests because we run the
validation code to ensure our images are always valid, and the
validation incorrectly flagged our images as invalid.

Fixes: ff477b3a7e85 ("vendor:update")
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -75,8 +75,15 @@ func (m *manifest) validate(w walker) error {
 		return errors.Wrap(err, "config validation failed")
 	}
 
+	validLayerMediaTypes := []string{
+		v1.MediaTypeImageLayer,
+		v1.MediaTypeImageLayerGzip,
+		v1.MediaTypeImageLayerNonDistributable,
+		v1.MediaTypeImageLayerNonDistributableGzip,
+	}
+
 	for _, d := range m.Layers {
-		if err := d.validate(w, []string{v1.MediaTypeImageLayer}); err != nil {
+		if err := d.validate(w, validLayerMediaTypes); err != nil {
 			return errors.Wrap(err, "layer validation failed")
 		}
 	}


cyphar:
This is also broken inside the image specification, which I'm fixing separately...

cyphar:
Can we please be more careful next time?

runcom:
This is a dupe of https://github.com/opencontainers/image-tools/pull/108 (though, this PR is correct...)

vbatts:
LGTM

image/manifest.go
```diff
@@ -75,8 +75,15 @@ func (m *manifest) validate(w walker) error {
 		return errors.Wrap(err, "config validation failed")
 	}
 
+	validLayerMediaTypes := []string{
```
wking: Is there a reason to make this private?  If a third party wants to support an extension layer type (which we [explicitly allow][1]), that would give them a way to do it.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/manifest.md#L62

wking:
Taking a step back, isn't this [already handled by image-spec][1] (which is was why it was [broken in two places][2])?  The DRY path forward seems like it would be dropping this image-tools code completely and using the image-spec validation instead.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/validator.go#L112-L117
[2]: https://github.com/opencontainers/image-tools/pull/111#issuecomment-278360889

wking:
image-spec calls that descriptor-type check (see [here][1] and [here][2]), so [`MediaTypeManifest.Validate`][3] will trigger it.  And image-tools [already][4] [calls][5] `MediaTypeManifest.Validate`.  I think the local manifest validation [6] should be “call `MediaTypeManifest.Validate` and (optionally?) recurse through descendants (because the image-spec validation is intra-blob).  Maintaining parallel intra-blob validation here seems like a waste of effort.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/validator.go#L55-L63
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/validator.go#L36
[3]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/schema.go#L26
[4]: https://github.com/opencontainers/image-tools/blob/f3b6fb91dff8d1434a63db19b76b08bf6d0fc257/image/manifest.go#L54
[5]: https://github.com/opencontainers/image-tools/blob/78c9ddb6228c858b37d1e2a7228f0af5d7d9dee2/cmd/oci-image-validate/main.go#L182
[6]: https://github.com/opencontainers/image-tools/blob/f3b6fb91dff8d1434a63db19b76b08bf6d0fc257/image/manifest.go#L73

xiekeyang:
Refer to https://github.com/opencontainers/image-tools/pull/62#issuecomment-263528341. This had better to be done in image spec repo.
cc @stevvooe WDYT?

cyphar:
@wking Sure, but at the moment that code doesn't actually return an error if the validation fails.

wking:
On Thu, Feb 09, 2017 at 12:43:56AM -0800, Aleksa Sarai wrote:
> Sure, but at the moment that code doesn't actually return an error
> if the validation fails.

That's because an unknown type is a SHOULD violation [1], not a MUST
violation, and the image-spec maintainers [2] and image-tools
maintainers (#66) haven't decided how they want to handle those yet.
Assuming we only error on MUST violations (until folks decide how to
handle SHOULD violations), there's no reason to error-out on unknown
types here, so using the image-spec code is fine.

Layer *unpacking*, on the other hand, should be erroring out on
unknown media types (that's violating the ‘unpackable’ condition I
floated in #66).  We'll want code around here [3], and that code will
have to start checking media types anyway now that both compressed and
uncompressed layers MUST be supported [4].

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/manifest.md#L62
[2]: https://github.com/opencontainers/image-spec/pull/491#discussion_r99456477
[3]: https://github.com/opencontainers/image-tools/blob/9386a7f89afec9215762e592aa054b7834c89aa4/image/manifest.go#L105-L106
[4]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/manifest.md#L55

stevvooe:
LGTM

coolljt0725:
LGTM

PR #113, opened 2017-02-10
Makefile: Add $(TOOLS) to .PHONY

q384566678:
Modify the command can be reused, the following problems will not occur：
`make: Nothing to be done for 'tools'.`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,7 @@ clean:
 .PHONY: \
 	all \
 	tools \
+	$(TOOLS) \
 	man \
 	install \
 	uninstall \


xiekeyang:
Is it necessary?

q384566678:
@xiekeyang  When you change the contents of the code, you want to re-generate the executable command.You must first execute `make clean` and then execute `make tools`, which is very troublesome.After this modification can be implemented directly `make tools`.

Makefile
```diff
@@ -100,8 +100,8 @@ clean:
 
 .PHONY: \
 	all \
-	tools \
-	man \
+	$(TOOLS) \
+	$(MAN) \
```
wking: You don't want to make this change.  The [`tools` target][1] is phony (building it does not create a file called `tools`), but the [`$(TOOLS)` targets][2] are real (building them creates the expected `oci-*` files).  Same for `man` (phony) and `$(MAN)` (not phony).

[1]: https://github.com/opencontainers/image-tools/blob/a358e03fde4e3628bf9fb7656bf643b63f975636/Makefile#L32
[2]: https://github.com/opencontainers/image-tools/blob/a358e03fde4e3628bf9fb7656bf643b63f975636/Makefile#L38-L39

wking: > [When you change the contents of the code, you want to re-generate the executable command.][1]

This is because we do not correctly list the dependencies for the `$(TOOLS)` (since calculating which Go source they depend on is not worth the trouble).  The fix should be to add `$(TOOLS)` to `.PHONY` and always rebuild (but don't remove `tools`).  The man-pages, on the other hand, *do* [correctly list their dependencies][3], so you don't need to add `$(MAN)`.

[1]: https://github.com/opencontainers/image-tools/pull/113#issuecomment-278872090
[2]: https://github.com/opencontainers/image-tools/blob/a358e03fde4e3628bf9fb7656bf643b63f975636/Makefile#L38-L39
[3]: https://github.com/opencontainers/image-tools/blob/a358e03fde4e3628bf9fb7656bf643b63f975636/Makefile#L42

q384566678: But if you don't add `$(MAN)`,you can't re-create the flie.

> make:Nothing to be done for 'man'.

wking: > But if you don't add $(MAN),you can't re-create the flie.

Sure you can, it just only rebuilds the man pages when it needs to:

```
$ git checkout origin/master
$ git describe --always
a358e03
$ make man
go-md2man -in "cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md" -out "oci-create-runtime-bundle.1"
go-md2man -in "cmd/oci-image-validate/oci-image-validate.1.md" -out "oci-image-validate.1"
go-md2man -in "cmd/oci-unpack/oci-unpack.1.md" -out "oci-unpack.1"
$ make man
make: Nothing to be done for 'man'.
$ touch cmd/oci-image-validate/oci-image-validate.1.md 
$ make man
go-md2man -in "cmd/oci-image-validate/oci-image-validate.1.md" -out "oci-image-validate.1"
```

q384566678: updated.

wking:
The diff in 71dfadbe looks good to me, but the commit message is not
very helpful.  The summary should be:

  Makefile: Add $(TOOLS) to .PHONY

or something like that.  And a commit message explaining that this is
because we don't tell make which files the Go targets depend on [1]
might help orient future readers who are trying to understand the
commit.

[1]: https://github.com/opencontainers/image-tools/pull/113#discussion_r100580795

xiekeyang:
LGTM

cyphar:
LGTM

coolljt0725:
LGTM

PR #116, opened 2017-02-16
Bump version for v0.1.0

xiekeyang:
There are merged 219 change-sets since the beginning and cleared the blockers for v0.1.0.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -25,7 +25,7 @@ const (
 	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = ""
+	VersionDev = "-dev"
 )
 
 // Version is the specification version that the package types support.
--- a/version/version.go
+++ b/version/version.go
@@ -20,12 +20,12 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 0
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 0
+	VersionMinor = 1
 	// VersionPatch is for backwards-compatible bug fixes
 	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "-dev"
+	VersionDev = ""
 )
 
 // Version is the specification version that the package types support.


xiekeyang:
Formerly version of image-tools followed up image-spec.
Here I present a new `version.go`, including the individual version of image-tools.

xiekeyang:
Change version to v0.1.0

xiekeyang:
@opencontainers/image-tools-maintainers We have got 7 LGTM votes for releasing v0.1.0. So PTAL thais PR. 
In this PR I set up new file for individual image-tools version, without relationship to image-spec. In future that image-tools supports all stuff in image-spec, then I think we might need to claim a compatible series of image-spec version, via `image-tools -v` command.

cmd/oci-image-validate/main.go
```diff
@@ -104,7 +104,7 @@ func newValidateCmd(stdout, stderr *log.Logger) *cobra.Command {
 func (v *validateCmd) Run(cmd *cobra.Command, args []string) {
 	if v.version {
 		v.stdout.Printf("commit: %s", gitCommit)
-		v.stdout.Printf("spec: %s", specs.Version)
+		v.stdout.Printf("spec: %s", version.Version)
```
cyphar: This isn't correct. We should output separate lines for what `spec` versions we support, and the version of the actual tool.

cyphar: Also, if we need to implement better version code we really should've gotten that merged _before_ we had a vote. The vote is "this commit should be tagged", not "this PR should be merged in whatever state it is when someone clicks the merge button".

xiekeyang:
> This isn't correct. We should output separate lines for what spec versions we support, and the version of the actual tool.

It is separated line. But what mean actual tool version?

cyphar:
The version of this tool is `v0.1.0`. It supports `v1.0.0-rc4` of the spec. This PR is removing the latter information and saying that the tool supports `0.1.0` of the spec (which is incorrect). This is what the output should look like:

```
% <blah> --version
<blah> version 0.1.0
commit: <some hash>
spec: 1.0.0-rc4
```

Or something like that.

xiekeyang:
> Also, if we need to implement better version code we really should've gotten that merged before we had a vote. The vote is "this commit should be tagged", not "this PR should be merged in whatever state it is when someone clicks the merge button".

I did put up this PR, and then mail for voting. Or you have other proposal to correct this?

cyphar:
@xiekeyang My concern is that you're adding new code for something that should just be a `VERSION` file bump. I mean, ultimately, I'd be willing to overlook it since we just didn't have the necessary code before (and I don't really want to go through another vote for no good reason). But I'm still a bit iffy about it.

xiekeyang:
@cyphar 

> The version of this tool is v0.1.0. It supports v1.0.0-rc4 of the spec. This PR is removing the latter information and saying that the tool supports 0.1.0 of the spec (which is incorrect). This is what the output should look like:

> % <blah> --version
> <blah> version 0.1.0
> commit: <some hash>
> spec: 1.0.0-rc4
> Or something like that.

Thanks, I should have keep spec version, and add tools version individual. Now I update commit to https://github.com/opencontainers/image-tools/pull/116/commits/704b914dd9d6c96070403f491a90167c9ba88aec. The version shows as:

```
$ ./xxx -v
commit: 98df00700ce69a83cde7a3707f3d5fab9fb86aa2
version: 0.1.0
spec: 1.0.0-rc4
```

PTAL, thanks.

xiekeyang:
@cyphar Maybe I should not publish the vote result in this PR, and reback the comment, which bring out confusion. But if the bump is ready? I think after it is merged, then I should tag and release v0.1.0 package?
cc @opencontainers/image-tools-maintainers

cyphar:
@xiekeyang For future reference, this is how the release process for OCI projects works:

1. You create a PR which has the list of issues and has two commits -- one which does the version bump and the other which switches it back to `vX.X.X+dev`. The commits should _only_ be those changes. An example of this is opencontainers/image-spec#566.

2. You send an email to the ML to call for a vote on the commit hash which contains the version bump (`5faaada` in the case of image-spec 1.0.0-rc5).

3. People vote.

4. Once the vote passes, we merge the PR and then push a signed git tag (which tags the commit hash that was voted on) to make the release.

The reason why it's done this way is so that people can be certain what they're voting on. As it stands, the vote we had references a commit which doesn't exist anymore -- which is why the commit that is being voted on should **not** contain code.

Now, I'm of half a mind that we break out the versioning changes into a separate pull request, get that merged and then have another vote about the `image-tools` version with a properly set up PR. But that's just my opinion and I'd like to know what the other @opencontainers/image-tools-maintainers have to say.

wking:
On Wed, Feb 22, 2017 at 11:26:13AM -0800, Aleksa Sarai wrote:
> 1. You create a PR which has the list of issues and has two commits
> -- one which does the version bump and the other which switches it
> back to `vX.X.X+dev`. The commits should _only_ be those changes. An
> example of this is opencontainers/image-spec#566.

You may want three commits, the two you mention here and an earlier
commit updating your release notes.  For example, see [1].  Having a
ChangeLog entry makes it easier for maintainers to see what they're
voting on, and makes it easy for other consumers to migate to the new
release.

There was some previous discussion about requiring ChangeLogs for OCI
releases [2], but as far as I'm aware attempts to define release
criteria have stalled out.  The Core Infrastructure Initiative Best
Practices Badge requires human-oriented release notes of some form
[3], and gives projects a few ways to meet that criterion, a ChangeLog
being one of them.  I expect OCI projects that don't already supply
release notes will begin adding them as they work on CII badging.

Cheers,
Trevor

[1]: https://github.com/opencontainers/runtime-spec/pull/672/commits
[2]: https://github.com/opencontainers/tob/pull/15#issuecomment-225988785
[3]: https://github.com/linuxfoundation/cii-best-practices-badge/blob/5a0a7d0effe1fd92f5313142f918a0de3a8d241f/doc/criteria.md#change-control

-- 
This email may be signed or encrypted with GnuPG (http://www.gnupg.org).
For more information, see http://en.wikipedia.org/wiki/Pretty_Good_Privacy

xiekeyang:
@cyphar Thank lots! I update it on https://github.com/opencontainers/image-tools/pull/116/commits/0cecf6b5442efa34fa4afa82f0726602559b681b and https://github.com/opencontainers/image-tools/pull/116/commits/69a24019be483718332e15ffba38782aaffbd94a.
While I still have to contain some code in https://github.com/opencontainers/image-tools/pull/116/commits/0cecf6b5442efa34fa4afa82f0726602559b681b, which is setting up image-tools version. If I separate it to individual commit, it is hard to write the right version. I think it should be OK for first version bump.

@wking As to add changelog, I didn't find it in similar cli tools repo as runtime-tools or docker/containerd. I think it might be not a well considered opinion yet, and I'd like to listen to opinions from @opencontainers/image-tools-maintainers.

cyphar:
@xiekeyang `umoci` has one, as do an incredibly large number of other projects (all GNU projects for example). And we should include a changelog (as someone who does packaging for a distribution, changelogs are really helpful).

But a changelog for 500 commits where we don't have a previous version isn't really helpful IMO.

cyphar:
@wking We're already having enough troubles bumping a version. Please let's not pile more things on.

xiekeyang:
@cyphar I agree the helpful changelog. But I honestly don't like to include it in this PR. Now I concern if this go forward on right direction.

wking:
On Wed, Feb 22, 2017 at 08:12:33PM -0800, Aleksa Sarai wrote:
> But a changelog for 500 commits where we don't have a previous
> version isn't really helpful IMO.

I agree that you probably don't need a ChangeLog for the first
release, since nobody is upgrading from a previous release.  But after
that, I think ChangeLogs are helpful.  Even if there are 500 commits.
Compiling a 500 commit ChangeLog entry would not be fun, but it is
much more efficient to do it once as a community than it would be to
have every consumer dig through those 500 commits on their own to see
what changed.

And if compiling a ChangeLog is so much trouble that it makes it hard
to cut a release, then I think you need to cut more frequent releases
and/or give up on CII badging.  Certainly don't stop cutting releases,
since that only makes it harder to figure out what changed.

xiekeyang:
@vbatts @cyphar @stevvooe 
If this bump PR is mostly ready, or still exist fault? Please point out directly if you find. Thank lots!

cyphar:
LGTM, though I'm still quite frustrated that we had a vote which was not helpful (also exposing the version through an API is not necessary -- but we can fix that later). Let's just please get a version out so we can stop being blocked on this and start fixing everything else in `image-tools` that needs to get fixed.

cyphar:
[ And for some reason PullApprove is still broken and doesn't recognise I'm a maintainer... ]

xiekeyang:
@cyphar Thanks for you explain and excuse me that I want to comply with the process of bump and vote like image-spec, but I didn't put enough attention to the potential conflict between them.

About 4 steps you mentioned, they are what I think as same as in image-spec, and had tried to carry out. But for vote mail on v0.1.0 release, I think we have achieve the agreement to release tag including which commits, except bump commit. We note that vote ML (both image-spec and image-tools) announces the bump PR id, not the final commit id, (I think) that because it might need fix more. So, can we accept that after this bump PR is landed, then I release the v0.1.0 tag, without more vote?

I'm not sure if I fully catch the right processing. Of course if necessary, I'd have to send a new mail, but please let me clear the needed supplement in the mail firstly, to avoid introduce more problems. To be honest, now I'm still a little confused if fixing this bump PR invalidates the previous vote result.

cc @opencontainers/image-tools-maintainers

cyphar:
@xiekeyang 

> We note that vote ML (both image-spec and image-tools) announces the bump PR id, not the final commit id, (I think) that because it might need fix more.

This is not correct. ML votes are on the **commit to be tagged**, *not* the pull request. This is why the subject of all vote emails is of [the format](https://github.com/opencontainers/project-template/blob/master/GOVERNANCE.md#subject-templates) `[runtime-spec VOTE]: Tag 0647920 as 1.0.0-rc (closes 2016-06-03 20:00 UTC)`.

The reason it's done this way (as as opposed to voting on a PR) is because if someone rebases a PR and now the version will refer to a different codebase, it's difficult to see what the maintainers voted on. Voting on a commit ID means that everyone agrees what the history will look like and what will be tagged.

Now, this means that you cannot change the contents of the voted on commit (since the commit ID will change). This is also a good thing, because it also makes sure that everyone agrees what the code will look like once it's merged.

However, because you cannot change the commit, this means that you **should not be making any code changes in a version bump**. The reason for this should be obvious given the above comments. As an example, IMO this PR is currently invalidated because the commit that we voted on contained code changes that needed to be fixed.

> So, can we accept that after this bump PR is landed, then I release the v0.1.0 tag, without more vote?

*Sigh*. While I feel like this is a bit too much bureaucracy-for-bureaucracy's-sake, we really shouldn't get into the habit of tagging commits that we didn't agree on.

> Of course if necessary, I'd have to send a new mail, but please let me clear the needed supplement in the mail firstly, to avoid introduce more problems. To be honest, now I'm still a little confused if fixing this bump PR invalidates the previous vote result.

Unfortunately in my mind, it does invalidate the vote result (well unless you want to merge the old code which we voted on -- which was also incorrect). Sorry that I didn't check this PR properly before voting. Here's what we need to do in order for votes to go far more smoothly in future:

1. Take the version changes in this PR (the code changes to make the tooling display the version of the tools) and put them into a separate PR. Make the initial version `0.0.0-dev` (which shouldn't need a vote because it's a dummy value).

2. Create a new PR after the above one is merged with two commits:
  - Bump to `v0.1.0`. **this is the commit we vote on**
  - Back to development (bump to `v0.1.0+dev`).

3. Post a mail on the ML to re-do the vote with the above commit.

4. Merge this once we get the vote done.

xiekeyang:
As #122 landed, I update this patch. @cyphar @vbatts @wking  please take an overlook. If this is the right direction, I would put up another vote by ML for v0.1.0 release. Thank lots!

cyphar:
This looks fine. Send the email. :smile_cat:

cyphar:
LGTM. Vote was `+7 -0 #3`.

coolljt0725:
LGTM

vbatts:
LGTM

cyphar:
Tag pushed, release cut, binaries built and signed with my keys. https://github.com/opencontainers/image-tools/releases/tag/v0.1.0

PR #117, opened 2017-02-16
image: Consistent reader usage

wking:
Start distinguishing between `reader` (when you only need to `Get` CAS blobs) and `walker` (when you might also need to access mutable refs).  This lays the ground work for something like #5, but avoids the contentious invention of new interfaces by sticking to interfaces which have already landed in master.  Also removes a number of redundant `Get` re-implementations.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -18,10 +18,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
-	"os"
-	"path/filepath"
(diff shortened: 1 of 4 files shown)


q384566678:
@wking This PR and #143  have some similar effects. Can you rebase this PR? And then we'll talk about the desirability of these.

wking:
It looks like `reader` was removed in #136, so there's no longer a way to make this distinction with the interfaces defined by the current image-tools.

PR #119, opened 2017-02-17
ignore generated man files

Mashimiao:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /oci-create-runtime-bundle
 /oci-unpack
 /oci-image-validate
+
+*.1


cyphar:
LGTM

.gitignore
```diff
@@ -1,3 +1,6 @@
 /oci-create-runtime-bundle
 /oci-unpack
 /oci-image-validate
+
+*.1
+*.patch
```
wking: Who makes these `*.patch` files?

Mashimiao: we always work with patch, and easy to forget to remove them before adding. I think it'd better to ignore them.

wking: I use `*.patch` files in projects where contribution is by mailing patches to a mailing list.  But I personally don't use them much in push-based projects like those on GitHub.  If you generate `*.patch` files as part of your personal workflow with this project, I think you should add a local ignore rule (gitignore(5) gives `$HOME/.config/git/ignore` and `$GIT_DIR/info/exclude` as possible locations).  But if rules in the committed `.gitignore` are based on workflow idiosyncrasies, it becomes hard to maintain them.

Mashimiao: Fine, updated

wking:
1434e4c looks good to me.

caniszczyk:
@cyphar btw PullApprove should catch your LGTMs moving forward, it was a misconfiguration on my end

cyphar:
LGTM

Mashimiao:

coolljt0725:
LGTM

PR #120, opened 2017-02-20
 Makefile: suitable "default" target

q384566678:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ TOOLS := \
 	oci-unpack
 MAN := $(TOOLS:%=%.1)
 
-default: help
+default: all
 
 help:
 	@echo "Usage: make <target>"


Makefile
```diff
@@ -10,7 +10,7 @@ TOOLS := \
 	oci-unpack
 MAN := $(TOOLS:%=%.1)
 
-default: help
+default: check-license lint test
```
cyphar: Why is this the most obvious default target? Why not `all` or `tools`?

q384566678: updated

q384566678:
@xiekeyang @coolljt0725 @stevvooe @cyphar PTAL

cyphar:
LGTM

stevvooe:
LGTM

PR #122, opened 2017-02-27
bump dummy version

xiekeyang:
During bumping official version, it should not change any code. So, it
has to prepare the version variables, their functions and dummy version
(as 0.0.0-dev) into code. After that is landed, then official versions
can be released.


xiekeyang:
it is prerequisite of #116. As previous discussion, normal version bumping can't chang any code. So here I prepare the necessary code on image-tools repo.

coolljt0725:
LGTM

cyphar:
LGTM so that we can get `v0.1.0` out of the door. But we really shouldn't be exposing this in the same way as `image-spec` -- the reasoning why `image-spec` exposes versioning this way is different.

PR #124, opened 2017-03-16
Add package of Sirupsen/logrus to vendor

xiekeyang:
supplement of #103. Here we need fix 2:
1. golint failed upper that go version 1.7
2. add logrus package. Refer to https://github.com/opencontainers/image-tools/pull/103#issuecomment-286932965

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.6
   - 1.7
+  - 1.8
 
 sudo: required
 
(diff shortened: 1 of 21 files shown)


xiekeyang:
I find `github.com/vbatts/git-validation` can help install non-existent packages to local, such as logrus, that pass travis CI. But I think we had better to keep as previous rules, to add needed package to glide.yaml, and commit it with code under vendor, it is right?

stevvooe:
LGTM

coolljt0725:
LGTM

PR #125, opened 2017-03-16
fix bug of failure on make lint

xiekeyang:
It is because golang version than 1.7 prompts as when golint,
```
os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019) (staticcheck)
```


xiekeyang:
Duplicated to #124

PR #129, opened 2017-03-22
man: Use OCI-IMAGE-TOOL-${COMMAND} in man headers

wking:
To match Git, which has headers like:

    GIT-INIT(1) Git Manual

in git-init(1).


q384566678:
LGTM

Mashimiao:
LGTM

PR #134, opened 2017-04-02
cmd/oci-image-tool/validate.go: create a logger if none exists.

erikh:
This was causing panics when validating an image with no refs, hitting this line of code, causing a panic on `out` being nil: https://github.com/opencontainers/image-tools/blob/master/image/image.go#L60

Thanks!

### DIFF
--- a/cmd/oci-image-tool/validate.go
+++ b/cmd/oci-image-tool/validate.go
@@ -96,6 +96,10 @@ func validatePath(name string) error {
 		}
 	}
 
+	if v.stdout == nil {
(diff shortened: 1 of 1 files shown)


vbatts:
LGTM

xiekeyang:
It seems duplicate to #46

vbatts:
@xiekeyang this is not nearly as extensive as #46, only fixing a panic. :smile:

coolljt0725:
LGTM

xiekeyang:
LGTM

PR #135, opened 2017-04-02
image/manifest.go: implement compression detection / mediatype validation

erikh:
Previously, this would effectively panic on any tar files because it was
expecting a gzip compressed file. This enforces that the `+gzip` media
type exists before enabling gzip decompression.

I also shimmed in some tests by appropriating some others.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -16,7 +16,9 @@ package image
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
+	"compress/bzip2"
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
@@ -27,6 +29,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/opencontainers/image-spec/schema"
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -120,8 +123,8 @@ func (m *manifest) unpack(w walker, dest string) (retErr error) {
 				return nil
 			}
 
-			if err := unpackLayer(dest, r); err != nil {
-				return errors.Wrap(err, "unpack: error extracting layer")
+			if err := unpackLayer(d.MediaType, path, dest, r); err != nil {
+				return errors.Wrap(err, "error unpack: extracting layer")
 			}
 
 			return errEOW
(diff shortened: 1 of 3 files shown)


cyphar:
The thing is, layers now can be either uncompressed or compressed. Maybe you should look at the auto-detect-compression code [here](https://github.com/containers/image/tree/master/pkg/compression)?

image/manifest.go
```diff
@@ -136,16 +136,22 @@ func (m *manifest) unpack(w walker, dest string) (retErr error) {
...
 
-func unpackLayer(dest string, r io.Reader) error {
+func unpackLayer(mediaType, dest string, r io.Reader) error {
 	entries := make(map[string]bool)
-	gz, err := gzip.NewReader(r)
-	if err != nil {
-		return errors.Wrap(err, "error creating gzip reader")
+
+	reader := r
+	if strings.HasSuffix(mediaType, "+gzip") {
```
cyphar: Oh sorry, you're already handling that. Oops.

erikh:
So the thing about this patch that's weird is that now the images I generate, which use tar files, are validating fine against oci-image-tool. I think I did something wrong on my end.

However this patch does validate the media type and switch the reader accordingly so it may be worth keeping. I'll leave it to you.

stevvooe:
@erikh I'd recommend auto-detecting compression to most compatible.

erikh:
@stevvooe not entirely sure what you mean by this, do you mean reading the file header?

stevvooe:
> do you mean reading the file header?

Yes.

erikh:
What's happening right now is that the media type is the source of truth
with this patch; I don't think this should change for validation purposes
at least. Perhaps we can appropriately detect and error on mismatched
compression with the media type?

This is the code used in `oci-image-tool validate` after all; it probably
shouldn't be too tolerant if it's to serve its purpose.

Happy to take it in any direction; just stating my opinion.

-Erik

stevvooe:
@erikh I'm just trying to provide a practical approach that has worked for us in employing unpacking for manifests. If you want to over-validate, go ahead, but its much more flexible to autodetect the compression in practice and there isn't really downside to it. Erroring out will just frustrate people that can do nothing about it. It is much better to emit a warning and unpack the layer. As long as the digest matches, it is okay.

For the most part, this will matter as people "shallow convert" existing images to the OCI formats.

Mind you, this is the only exception where I would say go against the media type. In almost all other cases, you should be strict.

erikh:
Fair enough. I will do that.

coolljt0725:
@erikh  conflict, needs a rebase

erikh:
It seems this needs an overhaul; I will do the work in the AM PST.

erikh:
PTAL and sorry for the wait.

This patch now:

* detect compression and appropriately adjusts the reader, warning if it doesn't match the media type
* proceeds to deal with the file anyway
* has some tests for testing invalid stuff
* also support for bzip compression in tests (I omitted xz but we could add that, I don't think it's in the spec?)

erikh:
I'll clean things up; needs a bit of work still. I'll comment when ready.

erikh:
I'll look later. Test is failing in CI but not locally.

erikh:
ok, this is green now. PTAL.

Note I had to import github.com/dsnet/compress/bzip2 to implement bzip2 compression. the hold-up was my pipe implementation that called bzip2, which works locally but doesn't seem to on travis for some reason.

vbatts:
LGTM

stevvooe:
LGTM

PR #136, opened 2017-04-04
image: accept arbitrary file streams

squeed:
This allows the image library to handle any ReadSeeker, enabling
on-the-fly uncompression.


image/image.go
```diff
@@ -31,16 +32,22 @@ func ValidateLayout(src string, refs []string, out *log.Logger) error {
...
 	defer f.Close()
 
-	return validate(newTarWalker(tarFile, f), refs, out)
+	return ValidateReader(f, refs, out)
+}
+
+// ValidateReader walks through the stream and validates the manifest
+// pointed to by the given refs or returns an error if the validation failed.
+// The stream must be a valid .tar file.
+func ValidateReader(r io.ReadSeeker, refs []string, out *log.Logger) error {
```
s-urbaniak: I'd even call this `Validate` as it is more general than the filename based one.

image/image.go
```diff
@@ -31,16 +32,22 @@ func ValidateLayout(src string, refs []string, out *log.Logger) error {
 	return validate(newPathWalker(src), refs, out)
 }
 
-// Validate walks through the given .tar file and validates the manifest
-// pointed to by the given refs or returns an error if the validation failed.
+// Validate opens the tar file given by the filename, then calls ValidateReader
 func Validate(tarFile string, refs []string, out *log.Logger) error {
```
s-urbaniak: I'd call this `ValidateFile(name string, refs []string, out *log.Logger`.

image/image.go
s-urbaniak: If we can't breakt the API, then I'd call this `ValidateReadSeeker` as it is more honest.

image/image.go
```diff
@@ -101,17 +108,22 @@ func UnpackLayout(src, dest, ref string) error {
...
 	defer f.Close()
 
-	return unpack(newTarWalker(tarFile, f), dest, ref)
+	return UnpackReader(f, dest, ref)
+}
+
+// UnpackReader walks through the tar stream and, using the layers specified in
+// the manifest pointed to by the given ref, unpacks all layers in the given
+// destination directory or returns an error if the unpacking failed.
+func UnpackReader(r io.ReadSeeker, dest, refName string) error {
```
s-urbaniak: same comments apply as for the `Validate*` methods.

image/image.go
squeed: I didn't want to break the public API - though I can if you like.

cyphar:
At the moment, the public API is going to have to be reworked anyway. Ultimately I don't _really_ care if we break it because:

a. We are not a 1.0 project.
b. There are a lot of things that need to be fixed with the current API that it doesn't make sense for us to mandate keeping a broken API (we're not kernel developers after all).

jonboulle:
agreed with aleksa - this looks reasonable for now. Needs signed-off-by at the least though.

squeed:
Updated the PR to rename methods according to @s-urbaniak's suggestions; I added a quick optimization while I was at it.

squeed:
Fixed `oci-image-tool create` while I was at it.

Not sure why "Signed-off-by" isn't being detected.

vbatts:
also, `make lint` has some feedback

squeed:
Whups, fixed, thanks.

vbatts:
Not sure about the @pullapprove failure on recognizing the DCO. It passed the `make .gitvalidation` check which also includes DCO. :-\

caniszczyk:
@vbatts is "Signed-off-by: Casey Callendrello <casey.Callendrello@coreos.com>" an email associated with @squeed's github account?

squeed:
Ah, probably just the case difference - I fixed the commit.

caniszczyk:
@squeed looks like that was it (sigh)

vbatts:
OHMAN
LGTM

image/image.go
```diff
@@ -143,17 +158,23 @@ func CreateRuntimeBundleLayout(src, dest, ref, root string) error {
 	return createRuntimeBundle(newPathWalker(src), dest, ref, root)
 }
 
-// CreateRuntimeBundle walks through the given .tar file and
-// creates an OCI runtime bundle in the given destination dest
-// or returns an error if the unpacking failed.
-func CreateRuntimeBundle(tarFile, dest, ref, root string) error {
+// CreateRuntimeBundleFile opens the file pointed by tarFile and calls
+// CreateRuntimeBundle.
```
xiekeyang: s/CreateRuntimeBundle/createRuntimeBundle ?

And, CreateRuntimeBundle which seems to be unused should be removed?

squeed: The purpose of this PR is to allow for arbitrary tar streams, making this more useful as a library.

While `CreateRuntimeBundle` isn't used by the cli tool, it's a good library function.

jonboulle:
:+1:

PR #137, opened 2017-04-06
cmd: Modify the assignment of parameters

q384566678:
This modification can avoid the absence of default values.

### DIFF
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -39,17 +39,10 @@ func createHandle(context *cli.Context) error {
 		return fmt.Errorf("both src and dest must be provided")
 	}
 
-	var v bundleCmd
-	if context.IsSet("type") {
-		v.typ = context.String("type")
-	}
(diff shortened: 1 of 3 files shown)


xiekeyang:
I'd rather to set value at statement of `v`, like,
```go
v := bundleCmd{root : rootfs, ref : ... }
```
And, I think `if` condition might be necessary for those who has no default value.

vbatts:
what happens if those variables are _not_ set? will it panic?

xiekeyang:
@vbatts 
> what happens if those variables are not set? will it panic?

it doesn't cause panic. The value should be empty for string type, `false` for bool type, and so on.

q384566678:
> And, I think if condition might be necessary for those who has no default value.

@xiekeyang The if statement performs only the assignment action, and the previous assignment has been repeated, so I do not think so.

xiekeyang:
LGTM

coolljt0725:
LGTM

PR #141, opened 2017-04-18
autodetect: add support for zip and rar archive type

Mashimiao:

### DIFF
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -26,6 +26,7 @@ import (
 var bundleTypes = []string{
 	image.TypeImageLayout,
 	image.TypeImage,
+	image.TypeImageZip,
 }
 
 type bundleCmd struct {
@@ -61,7 +62,7 @@ func createHandle(context *cli.Context) error {
 		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
 
 	case image.TypeImageZip:
-		err = image.CreateRuntimeBundleZip(context.Args()[0], context.Args()[1], v.ref, v.root)
+		err = image.CreateRuntimeBundleZip(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
 
 	case image.TypeImage:
 		err = image.CreateRuntimeBundleFile(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -60,6 +60,9 @@ func createHandle(context *cli.Context) error {
 	case image.TypeImageLayout:
 		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
 
+	case image.TypeImageZip:
+		err = image.CreateRuntimeBundleZip(context.Args()[0], context.Args()[1], v.ref, v.root)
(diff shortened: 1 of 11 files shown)


Mashimiao:
ping @opencontainers/image-tools-maintainers   any comments?

q384566678:
I think you need to add `TypeImageZip` in `validateTypes` and other places. Also need to modify the man file.

q384566678:
And please rebase and modify.

Mashimiao:
rebased

q384566678:
LGTM

Mashimiao:
ping @coolljt0725 @xiekeyang

q384566678:
LGTM

q384566678:

coolljt0725:
This is not work for me. 
```
$ oci-image-tool validate busybox.zip
1 errors detected:
busybox.zip: validation failed: file does not exist
```
I debugged and guess it should be something wrong with `zipWalker`   (https://github.com/opencontainers/image-tools/pull/141/files#diff-52676de86adecced88d6dc21eb58fbcaR222), the `get` method is same with the `tarWalker` here, but I think it should be different, because when open a tar archive with vim it shows like this
```
" tar.vim version v29
" Browsing tarfile /home/lei/images/busybox.tar
" Select a file with cursor and press ENTER

busybox/
busybox/oci-layout
busybox/index.json
busybox/blobs/
busybox/blobs/sha256/
busybox/blobs/sha256/8ac8bfaff55af948c796026ee867448c5b5b5d9dd3549f4006d9759b25d4a893
busybox/blobs/sha256/8baf43d43a34a0e6649c254b0200c2406fc40a501a852ba51a86ac3672dc0441
busybox/blobs/sha256/40a114053d955a2b80ee2cf6e13410b28b59594ceee9036b41e12c42d3e16615

```
but  open a zip archive with vim, it shows like this
```
" zip.vim version v27
" Browsing zipfile /home/lei/images/busybox.zip
" Select a file with cursor and press ENTER

blobs/
index.json
oci-layout

```
I think that's why this can't work?

Does this work for you ? or Am I missing something?

Mashimiao:
I think that maybe because you didn't put oci-layout, blobs and index.json into the root of the archive. A valid archive of image should have those at the root of the archive, not nested within a top-level directory as busybox.

q384566678:
It works for me:

```
➜  oci-image-tool validate ubuntu.zip 
ubuntu.zip: OK
Validation succeeded
```

I'm looking for the reason for your mistake.

coolljt0725:
sorry, It seems something wrong with my zip archive, it works with another zip archive
```
$ oci-image-tool validate busybox.zip
busybox.zip: OK
Validation succeeded

```

coolljt0725:
Can we  avoid adding a new `TypeImageZip` type and just use `TypeImage`, let the `TypeImage` to handle the different archive format? An zip format archive is also a file for image tool.  And in this way, I think we can reduce some duplication code. WDYT?

Mashimiao:
Letting TypeImage to handle different archive is a good option. But I think we have to do a lot changes to reach that.
Based on current structure, add TypeImageZip is the best choice, and rebuild the structure has less to do with the subject - adding support zip and rar.
So, how about making a follow-up PR to implement that after this PR landed? I can submit such PR later.

coolljt0725:
@Mashimiao  ok.  LGTM wait for other's opinion  @opencontainers/image-tools-maintainers

xiekeyang:
LGTM

PR #144, opened 2017-05-08
make the image-tool work with -rc5

q384566678:
This is a #121 follow-up PR, make the image-tool work with -rc5. Sorry so late update.

### DIFF
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -23,17 +23,14 @@ import (
 	"strings"
 
 	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
-type descriptor struct {
-	MediaType string `json:"mediaType"`
-	Digest    string `json:"digest"`
-	Size      int64  `json:"size"`
-}
+type descriptor v1.Descriptor
 
 func (d *descriptor) algo() string {
-	pts := strings.SplitN(d.Digest, ":", 2)
+	pts := strings.SplitN(string(d.Digest), ":", 2)
 	if len(pts) != 2 {
 		return ""
 	}
@@ -41,7 +38,7 @@ func (d *descriptor) algo() string {
 }
 
 func (d *descriptor) hash() string {
-	pts := strings.SplitN(d.Digest, ":", 2)
+	pts := strings.SplitN(string(d.Digest), ":", 2)
 	if len(pts) != 2 {
 		return ""
 	}
@@ -50,17 +47,22 @@ func (d *descriptor) hash() string {
 
 func listReferences(w walker) (map[string]*descriptor, error) {
 	refs := make(map[string]*descriptor)
+	var index v1.ImageIndex
 
 	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
-		if info.IsDir() || !strings.HasPrefix(path, "refs") {
+		if info.IsDir() || filepath.Clean(path) != "index.json" {
 			return nil
 		}
 
-		var d descriptor
-		if err := json.NewDecoder(r).Decode(&d); err != nil {
+		if err := json.NewDecoder(r).Decode(&index); err != nil {
 			return err
 		}
-		refs[info.Name()] = &d
+
+		for i := 0; i < len(index.Manifests); i++ {
+			if index.Manifests[i].Descriptor.Annotations["org.opencontainers.ref.name"] != "" {
+				refs[index.Manifests[i].Descriptor.Annotations["org.opencontainers.ref.name"]] = (*descriptor)(&index.Manifests[i].Descriptor)
+			}
+		}
 
 		return nil
 	}); err != nil {
@@ -71,21 +73,28 @@ func listReferences(w walker) (map[string]*descriptor, error) {
 
 func findDescriptor(w walker, name string) (*descriptor, error) {
 	var d descriptor
-	dpath := filepath.Join("refs", name)
+	var index v1.ImageIndex
 
 	switch err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
-		if info.IsDir() || filepath.Clean(path) != dpath {
+		if info.IsDir() || filepath.Clean(path) != "index.json" {
 			return nil
 		}
 
-		if err := json.NewDecoder(r).Decode(&d); err != nil {
+		if err := json.NewDecoder(r).Decode(&index); err != nil {
 			return err
 		}
 
-		return errEOW
+		for i := 0; i < len(index.Manifests); i++ {
+			if index.Manifests[i].Descriptor.Annotations["org.opencontainers.ref.name"] == name {
+				d = (descriptor)(index.Manifests[i].Descriptor)
+				return errEOW
+			}
+		}
+
+		return nil
 	}); err {
 	case nil:
-		return nil, fmt.Errorf("%s: descriptor not found", dpath)
+		return nil, fmt.Errorf("index.json: descriptor not found")
 	case errEOW:
 		return &d, nil
 	default:
@@ -105,7 +114,7 @@ func (d *descriptor) validate(w walker, mts []string) error {
 		return fmt.Errorf("invalid descriptor MediaType %q", d.MediaType)
 	}
 
-	parsed, err := digest.Parse(d.Digest)
+	parsed, err := digest.Parse(string(d.Digest))
 	if err != nil {
 		return err
 	}
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -20,33 +20,13 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
-	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
-type descriptor v1.Descriptor
-
-func (d *descriptor) algo() string {
-	pts := strings.SplitN(string(d.Digest), ":", 2)
-	if len(pts) != 2 {
-		return ""
-	}
-	return pts[0]
-}
-
-func (d *descriptor) hash() string {
-	pts := strings.SplitN(string(d.Digest), ":", 2)
-	if len(pts) != 2 {
-		return ""
-	}
-	return pts[1]
-}
-
-func listReferences(w walker) (map[string]*descriptor, error) {
-	refs := make(map[string]*descriptor)
+func listReferences(w walker) (map[string]*v1.Descriptor, error) {
+	refs := make(map[string]*v1.Descriptor)
 	var index v1.ImageIndex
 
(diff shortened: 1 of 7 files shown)


image/descriptor.go
```diff
@@ -48,19 +50,36 @@ func (d *descriptor) hash() string {
 	return pts[1]
 }
 
+// Index implementation
+type Index v1.ImageIndex
```
vbatts: This is not really needed. Is there a particular reason to make a local type for this struct? I could end up complicating golang type assertions

image/descriptor.go
```diff
@@ -48,19 +50,36 @@ func (d *descriptor) hash() string {
...
 
 	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
-		if info.IsDir() || !strings.HasPrefix(path, "refs") {
+		if info.IsDir() || filepath.Clean(path) != "index.json" {
 			return nil
 		}
 
-		var d descriptor
-		if err := json.NewDecoder(r).Decode(&d); err != nil {
+		buf, err := ioutil.ReadAll(r)
```
vbatts: doing a `.ReadAll` is a bit more expensive, where the previous version is a streaming decoder. I'd prefer a streaming decoder.

image/descriptor.go
```diff
@@ -71,21 +90,35 @@ func listReferences(w walker) (map[string]*descriptor, error) {
...
+	var index Index
 
 	switch err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
-		if info.IsDir() || filepath.Clean(path) != dpath {
+		if info.IsDir() || filepath.Clean(path) != "index.json" {
 			return nil
 		}
 
-		if err := json.NewDecoder(r).Decode(&d); err != nil {
+		buf, err := ioutil.ReadAll(r)
```
vbatts: same

image/descriptor.go
q384566678: I need this struct to carry the contents of `index.json`.

image/descriptor.go
q384566678: updated. PTAL.

image/descriptor.go
cyphar: But why not just use `v1.ImageIndex`? You're not defining any methods so there should be no need to add this extra level of type indirection.

q384566678: Oh I see, updated, thanks.

runcom:
@cyphar PTAL (need this for skopeo tests in https://github.com/projectatomic/skopeo/pull/313)

q384566678:
What do you think of this PR? Some of the improvements in `image-tools` need to be done on this basis. @vbatts @runcom @cyphar @stevvooe @coolljt0725 @xiekeyang PTAL

image/descriptor.go
```diff
@@ -50,17 +51,26 @@ func (d *descriptor) hash() string {
...
 		}
-		refs[info.Name()] = &d
+
+		var d descriptor
+		for i := 0; i < len(index.Manifests); i++ {
+			if index.Manifests[i].Descriptor.Annotations["org.opencontainers.ref.name"] != "" {
+				d.MediaType = index.Manifests[i].Descriptor.MediaType
+				d.Digest = string(index.Manifests[i].Descriptor.Digest)
+				d.Size = index.Manifests[i].Descriptor.Size
+				refs[index.Manifests[i].Descriptor.Annotations["org.opencontainers.ref.name"]] = &d
```
cyphar: This will completely flatten the set of references (silently ignoring duplicates). I've spent quite a while trying to come up with a sane API for `umoci` and I think I've decided the only way to implement this is to remove the concept of a reference in the lower-level libraries and have it as a higher-level construct.

In `umoci` this split is quite easy to do `oci/cas` vs `oci/casext` which adds implementation-agnostic extensions to `oci/cas`. Maybe that's something you can do so you can remove the old concepts of a reference being a simple mapping.

runcom: yeah, and just append here.

cyphar: @stevvooe ... because first-class references aren't a property of the underlying storage anymore ...

Whatever, I'm not going to get into another argument over this.

q384566678: I try to use `type descriptor v1.Descriptor` instead of descriptor, but the following error occurs:

> /tmp/go-build341304265/github.com/opencontainers/image-tools/image/_test/_obj_test/descriptor.go:47:cannot use d.Digest (type digest.Digest) as type string in argument to strings.SplitN
/tmp/go-build341304265/github.com/opencontainers/image-tools/image/_test/_obj_test/descriptor.go:81: cannot use &index.Manifests[i].Descriptor (type *v1.Descriptor) as type *descriptor in assignment

What I am using is what I can think of.As for the kind of ideas you say I want to do, but can not solve the problem after the amendment.

stevvooe: > Whatever, I'm not going to get into another argument over this.

Please avoid the passive aggressive tone.

stevvooe: @q384566678 You need to convert to string: `dgst.String()` should do the trick. However, it also looks like this local descriptor type is no longer needed.

q384566678: @stevvooe Sorry to disturb, I have solved other problems, only the problem can not be resolved. 
> descriptor.go:64: cannot use index.Manifests[i].Descriptor (type v1.Descriptor) as type descriptor in assignment

Or is there a problem with my method. Can we discuss this question? If can't solve the problem I think I still use the method in https://github.com/opencontainers/image-tools/pull/144/commits/ea52a5986f7dcd7af63519784b60debf85ebddae.  （ I think this method is feasible.）
I hope to also try to promote the development of image-tool, but sometimes some technical problems need your help.Thanks.

stevvooe: @q384566678 The problem is that there is a local type `descriptor` and a common type, `v1.Descriptor`. You either need to convert the types or just use the `v1.Descriptor` type.

q384566678:
@runcom @stevvooe @vbatts @cyphar updated.PTAL

q384566678:

q384566678:
@opencontainers/image-tools-maintainers I think this can be merged.PTAL, thanks.

image/descriptor.go
```diff
@@ -23,25 +23,22 @@ import (
...
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
-type descriptor struct {
-	MediaType string `json:"mediaType"`
-	Digest    string `json:"digest"`
-	Size      int64  `json:"size"`
-}
+type descriptor v1.Descriptor
```
stevvooe: Why make this type definition? Just use the type.

image/descriptor.go
```diff
@@ -23,25 +23,22 @@ import (
...
 
 func (d *descriptor) algo() string {
```
stevvooe: Lost this code: this just go-digest to do this properly.

stevvooe:
@q384566678 Rather than defining a new descriptor type, just use the existing one.

image/descriptor.go
q384566678: If not, two descriptor cannot be assigned. （Type conversion is not ok. ）

> descriptor.go:67: cannot convert &index.Manifests[i].Descriptor (type *v1.Descriptor) to type *descriptor

If you want to use the original definition as you say, you can only assign each structure separately in the original way.

image/descriptor.go
q384566678: updated.

q384566678:
@stevvooe 
If not, two descriptor cannot be assigned. （Type conversion is not ok. ）

> descriptor.go:67: cannot convert &index.Manifests[i].Descriptor (type *v1.Descriptor) to type *descriptor

If you want to use the original definition as you say, you can only assign each structure separately in the original way:

> d.MediaType = index.Manifests[i].Descriptor.MediaType
  d.Digest = string(index.Manifests[i].Descriptor.Digest)
  d.Size = index.Manifests[i].Descriptor.Size

image/descriptor.go
stevvooe: Remove the type def please.

stevvooe:
@q384566678 The type definition is not needed. Remove the extra type and it will work. You may need to update a few to make it work.

image/descriptor.go
q384566678: updated,I used to define this type because of this function, 

> func (d *descriptor) validate(w walker, mts []string) error {

and now I changed it to

> func ValidateDescriptor(d *v1.Descriptor, w walker, mts []string) error {

PTAL

image/descriptor.go
```diff
@@ -69,31 +52,39 @@ func listReferences(w walker) (map[string]*descriptor, error) {
...
 	case errEOW:
 		return &d, nil
 	default:
 		return nil, err
 	}
 }
 
-func (d *descriptor) validate(w walker, mts []string) error {
+//ValidateDescriptor validate descriptor type
+func ValidateDescriptor(d *v1.Descriptor, w walker, mts []string) error {
```
stevvooe: No need to export this. It is use unexported interfaces, so it is pointless, anyways.

image/descriptor.go
```diff
@@ -105,7 +96,7 @@ func (d *descriptor) validate(w walker, mts []string) error {
 		return fmt.Errorf("invalid descriptor MediaType %q", d.MediaType)
 	}
 
-	parsed, err := digest.Parse(d.Digest)
+	parsed, err := digest.Parse(string(d.Digest))
```
stevvooe: No need to parse a digest that is already a digest. Just call `d.Digest.Validate()`.

q384566678: updated, PTAL, thanks.

image/image_test.go
```diff
@@ -291,41 +321,41 @@ func createTarBlob(name string, list []tarContent) error {
...
-func createHashedBlob(name string) (descriptor, error) {
+func createHashedBlob(name string) (v1.Descriptor, error) {
 	desc, err := newDescriptor(name)
 	if err != nil {
-		return descriptor{}, err
+		return v1.Descriptor{}, err
 	}
 
-	parsed, err := digest.Parse(desc.Digest)
+	parsed, err := digest.Parse(string(desc.Digest))
```
stevvooe: `desc.Digest.Validate()`

q384566678: updated, PTAL.

stevvooe:
LGTM

q384566678:
@xiekeyang @coolljt0725 @vbatts PTAL.

cyphar:
Do we want to add explicit validation for `index.json`?

LGTM otherwise, we can do that in a follow-up.

cyphar:
@q384566678 Can you answer this question please?

> Do we want to add explicit validation for index.json?

q384566678:
@cyphar I wrote a validation that contained the existence of index.json in #84  and the validation of index.json content in #51 (I will convert the validation of the manifest list to index.json.).

cyphar:
:+1:

PR #146, opened 2017-05-15
cmd: small fixes

q384566678:


vbatts:
not really needed. If a rename would be needed at all, i would expect it to be something like `createAction` or similar.

q384566678:
> not really needed. If a rename would be needed at all, i would expect it to be something like createAction or similar.

ok,I just want them to be in the same format.

Mashimiao:
LGTM

PR #149, opened 2017-05-31
image: modify to create bundlle

q384566678:
Not just linux can create runtime bundle. As for the three systems outside the os, I am not sure whether it can be discussed in the [runtime-tool#347.](https://github.com/opencontainers/runtime-tools/pull/347)


image/config.go
```diff
@@ -66,11 +66,19 @@ func findConfig(w walker, d *descriptor) (*config, error) {
...
 func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
-	if c.OS != "linux" {
+	var s specs.Spec
+
+	switch c.OS {
+	case "linux":
+		s.Linux = &specs.Linux{}
+	case "solaris":
+		s.Solaris = &specs.Solaris{}
+	case "winodws":
```
cyphar: `windows`

cyphar:
The biggest problem with the other platforms is extraction. How do you (sanely) extract layers on a Windows filesystem (there's a huge number of knobs in `tar` that don't make sense on Windows -- what should `major`, `minor`, `uid`, `gid` map to?) Solaris should be fine though.

q384566678:
> The biggest problem with the other platforms is extraction. How do you (sanely) extract layers on a Windows 

Uh, I don't know much about the problem you're talking about. If you have any good suggestions, please suggest, thank you.

q384566678:
ping @xiekeyang @coolljt0725 @vbatts @stevvooe PTAL

q384566678:
reping @coolljt0725 @stevvooe @cyphar

PR #150, opened 2017-06-02
vendor: update image-spec to rc6

q384566678:


q384566678:
@cyphar @stevvooe @vbatts @coolljt0725 @jonboulle PTAL.

q384566678:
@opencontainers/image-tools-maintainers I think I need to get some response.

coolljt0725:
should we make sure that `image-tool` work with rc5 well and then we release a `image-tool`, after that we update the image-spec to rc6 ?

q384566678:
> should we make sure that image-tool work with rc5 well and then we release a image-tool, after that we update the image-spec to rc6 ?

I think it is basically in line with rc5, only one validation of index, which I think will be more convenient after rc6.  I think some of the verification (including imageLayout) in the update to rc6 will be better implemented.

q384566678:
@coolljt0725 @xiekeyang @cyphar @vbatts @stevvooe `Image-tool` has been released v0.2.0 version, can be updated to rc6?

q384566678:
re-ping @opencontainers/image-tools-maintainers.

xiekeyang:
I think it is OK, what other maintainers think?

coolljt0725:
LGTM

stevvooe:
LGTM

PR #151, opened 2017-06-14
lint: update the check range

q384566678:


Mashimiao:
why do we need this?

q384566678:
Closed.

PR #154, opened 2017-06-28
image: bug fixes

q384566678:


Mashimiao:
I don't think this is a bug at all.

PR #157, opened 2017-07-06
image: avoid panic on nil fields when converting

q384566678:
When I execute the create command, there was an error, but there was no error when making the test. So need to improve the test content, and fix the error.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -73,8 +73,11 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 	var s specs.Spec
 	s.Version = specs.Version
 	// we should at least apply the default spec, otherwise this is totally useless
-	s.Process.Terminal = true
+	s.Root = &specs.Root{}
 	s.Root.Path = rootfs
+
+	s.Process = &specs.Process{}
+	s.Process.Terminal = true
 	s.Process.Cwd = "/"
 	if c.Config.WorkingDir != "" {
 		s.Process.Cwd = c.Config.WorkingDir
--- a/image/image_test.go
+++ b/image/image_test.go
@@ -187,12 +187,36 @@ type imageLayout struct {
 	tarList   []tarContent
 }
 
-func TestValidateLayout(t *testing.T) {
+func TestImageLayout(t *testing.T) {
 	root, err := ioutil.TempDir("", "oci-test")
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer os.RemoveAll(root)
+
+	dest1, err := ioutil.TempDir("", "dest1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dest1)
+
+	dest2, err := ioutil.TempDir("", "dest2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dest2)
+
+	dest3, err := ioutil.TempDir("", "dest3")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dest3)
+
+	dest4, err := ioutil.TempDir("", "dest4")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dest4)
 
 	il := imageLayout{
 		rootDir:   root,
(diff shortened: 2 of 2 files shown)


q384566678:

q384566678:
@xiekeyang @coolljt0725 @cyphar @vbatts  PTAL

cyphar:
What was the error? Can you include it in the commit message that fixes said error?

q384566678:
The following error occurred:
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x48ef43]
goroutine 19 [running]:
panic(0x808060, 0xc420010130)
	/home/zhouhao/go/src/runtime/panic.go:500 +0x1ae
testing.tRunner.func1(0xc4200932c0)
	/home/zhouhao/go/src/testing/testing.go:579 +0x474
panic(0x808060, 0xc420010130)
	/home/zhouhao/go/src/runtime/panic.go:458 +0x271
github.com/opencontainers/image-tools/image.(*config).runtimeSpec(0xc420676240, 0x85e3f6, 0x6, 0xc4201d3920, 0x1a, 0x0)
	github.com/opencontainers/image-tools/image/_test/_obj_test/config.go:94 +0x1b3
github.com/opencontainers/image-tools/image.createBundle(0xa2f7e0, 0xc4204fc090, 0xc420543b20, 0xc4200df2a0, 0x13, 0x85e3f6, 0x6, 0x0, 0x0)
	github.com/opencontainers/image-tools/image/_test/_obj_test/image.go:414 +0x318
github.com/opencontainers/image-tools/image.createRuntimeBundle(0xa2f7e0, 0xc4204fc090, 0xc4200df2a0, 0x13, 0x85e384, 0x6, 0x85e3f6, 0x6, 0x0, 0x0, ...)
	github.com/opencontainers/image-tools/image/_test/_obj_test/image.go:348 +0x9ce
github.com/opencontainers/image-tools/image.CreateRuntimeBundleLayout(0xc4200df180, 0x16, 0xc4200df2a0, 0x13, 0x85e384, 0x6, 0x85e3f6, 0x6, 0x0, 0x0, ...)

Because there is no non-empty check on the optional fields in config.go.

image/config.go
```diff
@@ -73,37 +73,41 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
...
 		}
+		s.Process.Env = append(s.Process.Env, c.Config.Env...)
+		s.Process.Args = append(s.Process.Args, c.Config.Entrypoint...)
+		s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
 
-		gid, err := strconv.Atoi(ug[1])
-		if err != nil {
-			return nil, errors.New("config.User: unsupported gid format")
+		if len(s.Process.Args) == 0 {
+			s.Process.Args = append(s.Process.Args, "sh")
```
stevvooe: Do we make the assumption that `sh` is always available in a container?

stevvooe:
A good title for this PR and the commit might be "image: avoid panic on nil fields when converting".

Also, please take the time to squash this into a single commit.

image/config.go
q384566678: Yes, `Args` has at least one value.

q384566678:
updated, PTAL.

image/config.go
stevvooe: That is not what I asked.

Not all containers are going to have a `sh` binary present. Adding this, when it might not even exist, will certainly result in failure. This is also never mentioned in the specification. In the case that no args are provided, it should just not be set.

stevvooe: Fixing this is outside the scope of this PR. I've filed https://github.com/opencontainers/image-tools/issues/166.

q384566678: @stevvooe If there are no other questions, can we merge this patch first and then discuss the `sh` problem? Because if you don't merge this, the `create` function won't work.

image/config.go
```diff
@@ -73,39 +73,43 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 	var s specs.Spec
 	s.Version = specs.Version
 	// we should at least apply the default spec, otherwise this is totally useless
-	s.Process.Terminal = true
-	s.Root.Path = rootfs
-	s.Process.Cwd = "/"
-	if c.Config.WorkingDir != "" {
-		s.Process.Cwd = c.Config.WorkingDir
+	if s.Root != nil {
```
Mashimiao: I think this is not a right fix. Default Root is nil in Spec, we should initialize it as s.Root= &specs.Root{},  also same to Process.

q384566678: updated, thanks.

stevvooe:
LGTM

coolljt0725:
LGTM

PR #158, opened 2017-07-07
Add go version requirement

q384566678:
Through [.travis.yml](https://github.com/opencontainers/image-tools/blob/master/.travis.yml#L3) know to use a go 1.7 and above, and go don't use GO15VENDOREXPERIMENT since 1.7


q384566678:

coolljt0725:
LGTM    image-tool use a new feature `io.SeekStart`(https://github.com/opencontainers/image-tools/blob/master/image/autodetect.go#L73) only available on 1.7.x and above,  I think make this requirement make sense

erikh:
archive/tar changes significantly in 1.9, which needed some fixes in docker
IIRC (detail here: https://github.com/moby/moby/pull/33935/). I don't know
if I'm really beaten by this myself, but it'd ease my mind a little if we
could at least limit it to 1.8 and up.

erikh:
Sorry I misread the point of this PR I think.

q384566678:
> Any reason not to take this up to 1.8 or 1.9?

I will go to find the relevant content, if necessary, I will be in the next PR to modify.

PR #161, opened 2017-07-19
Makefile: add gofmt and govet to test

q384566678:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,14 +59,17 @@ lint:
 	@echo "checking lint"
 	@./.tool/lint
 
-.PHONY: .gofmt .gotest
+.PHONY: .gofmt .govet .gotest
 
 PACKAGES = $(shell go list ./... | grep -v /vendor/)
-test: .gofmt .gotest
+test: .gofmt .govet .gotest
 
 FILES = $(shell find ./ -name *.go | grep -v vendor)
 .gofmt:
 	OUT=$$(gofmt -s -d $(FILES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
+
+.govet:
+	go vet -x $(PACKAGES)
 
 .gotest:
 	go test -v -race -cover $(PACKAGES)
--- a/Makefile
+++ b/Makefile
@@ -59,8 +59,17 @@ lint:
 	@echo "checking lint"
 	@./.tool/lint
 
-test:
-	go test -v -race -cover $(shell go list ./... | grep -v /vendor/)
+.PHONY: .gofmt .gotest
+
+PACKAGES = $(shell go list ./... | grep -v /vendor/)
+test: .gofmt .gotest
+
+FILES = $(shell find ./ -name *.go | grep -v vendor)
+.gofmt:
+	OUT=$$(gofmt -s -d $(FILES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
+
+.gotest:
+	go test -v -race -cover $(PACKAGES)
 
 
 ## this uses https://github.com/Masterminds/glide and https://github.com/sgotti/glide-vc


q384566678:

Makefile
```diff
@@ -59,8 +59,16 @@ lint:
 	@echo "checking lint"
 	@./.tool/lint
 
-test:
-	go test -v -race -cover $(shell go list ./... | grep -v /vendor/)
+.PHONY: .gofmt .gotest
+
+PACKAGES = $(shell go list ./... | grep -v /vendor/)
+test: .gofmt .gotest
```
stevvooe: Why would format be part of testing?

Makefile
```diff
@@ -59,8 +59,16 @@ lint:
...
+
+.gofmt:
+	OUT=$$(go fmt $(PACKAGES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
```
stevvooe: Use `gofmt -s -d`. The output as you have it won't be very useful.

Makefile
q384566678: I want to do `fmt` detection for the format, so I think it belongs to test, and I can also add `gofmt `commands in the makefile  to make formatting adjustments.

Makefile
```diff
@@ -59,8 +59,19 @@ lint:
...
 
-test:
-	go test -v -race -cover $(shell go list ./... | grep -v /vendor/)
+.PHONY: .gofmt .gotest .govet
+
+PACKAGES = $(shell go list ./... | grep -v /vendor/)
+test: .gofmt .govet .gotest
+
+.gofmt:
+	OUT=$$(go fmt $(PACKAGES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
```
stevvooe: This still isn't using `gofmt -d -s`. Please use the correct command.

q384566678: I tried the method you mentioned, but `gofmt` only supports absolute paths. My idea is to use `pwd` to get the absolute path of `image-tools`, but I don't know how to remove the `vendor` next .
Do you have any advice?

Mashimiao: how about like as following?
```
FILES = $(shell find ./ -name *.go | grep -v vendor)
OUT=$$(gofmt -s -d $(FILES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
```

q384566678: I think it's good, thanks.

q384566678:
ping @coolljt0725 @stevvooe

q384566678:
@xiekeyang @stevvooe @coolljt0725 PTAL

q384566678:
reping @opencontainers/image-tools-maintainers

q384566678:
ping @coolljt0725 @stevvooe @xiekeyang @vbatts

q384566678:
Bump

q384566678:
@stevvooe @coolljt0725 @xiekeyang @cyphar PTAL

Makefile
```diff
@@ -59,8 +59,21 @@ lint:
...
+FILES = $(shell find ./ -name *.go | grep -v vendor)
+.gofmt:
+	OUT=$$(gofmt -s -d $(FILES)); if test -n "$${OUT}"; then echo "$${OUT}" && exit 1; fi
+
+.govet:
+	go vet -x $(PACKAGES)
+
+UTDIRS = ./image/...
+.gotest:
+	go test -v -race -cover $(UTDIRS)
```
coolljt0725: why `go test ` only cover `UTDIRS = ./image/...` ? the old way is to cover `$(shell go list ./... | grep -v /vendor/)`?

q384566678: >?   	github.com/opencontainers/image-tools/cmd/oci-image-tool	[no test files]
ok  	github.com/opencontainers/image-tools/image	1.120s	coverage: 56.5% of statements
?   	github.com/opencontainers/image-tools/version	[no test files]

Because only the image has the test file, this can improve efficiency.

coolljt0725: But I think it's better to keep the old way, because we can't make sure we will not add test file in other dir in future.

q384566678: Fine, updated.

coolljt0725:
with this patch and I run `make test`, I can see
```
OUT=$(gofmt -s -d ./image/image_test.go ./image/descriptor.go ./image/config.go ./image/manifest_test.go ./image/index.go ./image/autodetect.go ./image/project.go ./image/layout.go ./image/doc.go ./image/manifest.go ./image/image.go ./image/walker.go ./cmd/oci-image-tool/create.go ./cmd/oci-image-tool/main.go ./cmd/oci-image-tool/unpack.go ./cmd/oci-image-tool/validate.go ./version/version.go); if test -n "${OUT}"; then echo "${OUT}" && exit 1; fi

```
Is this a expected output?

q384566678:
>Is this a expected output?

Yes.The purpose is to detect all the `.go` file.

coolljt0725:
IMO, we can just output the gofmt warning if there is and output nothing if there is no format  warning

q384566678:
I think it should be output, just like `test`, the process should be displayed.

Mashimiao:
LGTM

stevvooe:
LGTM

PR #162, opened 2017-07-19
image/image.go: cleanup and fix comment

Mashimiao:
CreateRuntimeBundle is not needed any more. we should cleanup it.

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -221,8 +221,9 @@ func CreateRuntimeBundleLayout(src, dest, ref, root, platform string) error {
 	return createRuntimeBundle(newPathWalker(src), dest, ref, root, platform)
 }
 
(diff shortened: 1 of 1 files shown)


q384566678:
This question is being discussed  [here](https://github.com/opencontainers/image-tools/pull/136#discussion-diff-110102565R161),  what about just fix comment?

Mashimiao:
Even though from the point of view of as a library, I still can't see any necessaries to keep it.

Mashimiao:

stevvooe:
@Mashimiao It's mostly best practice to have an `io.Reader` version, in case one wants to generate the tar file into something other than a file on disk.

Mashimiao:
Fine, that's not a big deal. Closing it.

PR #169, opened 2017-07-31
Modify the ref option

### LINKED ISSUE #164: How to handle multiple manifests with the same ref.name
When the implementation of `create` or `unpack`, if according to the input ref, in the `index.json` can be found in a number of correspondence, this situation should be how to deal with? Is it supposed to handle all? Or deal with one? I was wondering if it should be added in the specification to increase the limit on `org.opencontainers.ref.name`?
cc/@opencontainers/image-tools-maintainers

q384566678:
Increase the search conditions, you can better retrieve descriptor.
Resolved when the search is not unique or non-existent.
cc/ @opencontainers/image-tools-maintainers 

Fixes #164

### DIFF
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -31,7 +31,7 @@ var bundleTypes = []string{
 
 type bundleCmd struct {
 	typ      string // the type to bundle, can be empty string
-	ref      string
+	refs     []string
 	root     string
 	platform string
 }
@@ -43,9 +43,13 @@ func createHandle(context *cli.Context) error {
 
 	v := bundleCmd{
 		typ:      context.String("type"),
-		ref:      context.String("ref"),
+		refs:     context.StringSlice("ref"),
 		root:     context.String("rootfs"),
 		platform: context.String("platform"),
+	}
+
+	if len(v.refs) == 0 {
+		return fmt.Errorf("ref must be provided")
 	}
 
 	if v.typ == "" {
@@ -59,13 +63,13 @@ func createHandle(context *cli.Context) error {
 	var err error
 	switch v.typ {
 	case image.TypeImageLayout:
-		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
+		err = image.CreateRuntimeBundleLayout(context.Args()[0], context.Args()[1], v.root, v.platform, v.refs)
 
 	case image.TypeImageZip:
-		err = image.CreateRuntimeBundleZip(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
+		err = image.CreateRuntimeBundleZip(context.Args()[0], context.Args()[1], v.root, v.platform, v.refs)
 
 	case image.TypeImage:
-		err = image.CreateRuntimeBundleFile(context.Args()[0], context.Args()[1], v.ref, v.root, v.platform)
+		err = image.CreateRuntimeBundleFile(context.Args()[0], context.Args()[1], v.root, v.platform, v.refs)
 
 	default:
 		err = fmt.Errorf("cannot create %q", v.typ)
@@ -87,10 +91,9 @@ var createCommand = cli.Command{
 				strings.Join(bundleTypes, ","),
 			),
 		},
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name:  "ref",
-			Value: "v1.0",
-			Usage: "The ref pointing to the manifest of the OCI image. This must be present in the 'refs' subdirectory of the image.",
+			Usage: "A set of ref specify the search criteria for the validated reference, format is A=B. Only support 'name', 'platform.os' and 'digest' three cases.",
 		},
 		cli.StringFlag{
 			Name:  "rootfs",
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -27,8 +28,30 @@ import (
 
 const indexPath = "index.json"
 
-func listReferences(w walker) (map[string]*v1.Descriptor, error) {
-	refs := make(map[string]*v1.Descriptor)
+func listReferences(w walker) ([]v1.Descriptor, error) {
+	var descs []v1.Descriptor
+	var index v1.Index
+
+	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
+		if info.IsDir() || filepath.Clean(path) != indexPath {
+			return nil
+		}
+
+		if err := json.NewDecoder(r).Decode(&index); err != nil {
+			return err
+		}
+		descs = index.Manifests
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return descs, nil
+}
+
+func findDescriptor(w walker, names []string) ([]v1.Descriptor, error) {
+	var descs []v1.Descriptor
 	var index v1.Index
 
 	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
@@ -40,9 +63,34 @@ func listReferences(w walker) (map[string]*v1.Descriptor, error) {
 			return err
 		}
 
-		for i := 0; i < len(index.Manifests); i++ {
-			if index.Manifests[i].Annotations[v1.AnnotationRefName] != "" {
-				refs[index.Manifests[i].Annotations[v1.AnnotationRefName]] = &index.Manifests[i]
+		descs = index.Manifests
+		for _, name := range names {
+			argsParts := strings.Split(name, "=")
+			if len(argsParts) != 2 {
+				return fmt.Errorf("each ref must contain two parts")
+			}
+
+			switch argsParts[0] {
+			case "name":
+				for i := 0; i < len(descs); i++ {
+					if descs[i].Annotations[v1.AnnotationRefName] != argsParts[1] {
+						descs = append(descs[:i], descs[i+1:]...)
+					}
+				}
+			case "platform.os":
+				for i := 0; i < len(descs); i++ {
+					if descs[i].Platform != nil && index.Manifests[i].Platform.OS != argsParts[1] {
+						descs = append(descs[:i], descs[i+1:]...)
+					}
+				}
+			case "digest":
+				for i := 0; i < len(descs); i++ {
+					if string(descs[i].Digest) != argsParts[1] {
+						descs = append(descs[:i], descs[i+1:]...)
+					}
+				}
+			default:
+				return fmt.Errorf("criteria %q unimplemented", argsParts[0])
 			}
 		}
 
@@ -50,38 +98,14 @@ func listReferences(w walker) (map[string]*v1.Descriptor, error) {
 	}); err != nil {
 		return nil, err
 	}
-	return refs, nil
-}
 
-func findDescriptor(w walker, name string) (*v1.Descriptor, error) {
-	var d v1.Descriptor
-	var index v1.Index
-
-	switch err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
-		if info.IsDir() || filepath.Clean(path) != indexPath {
-			return nil
-		}
-
-		if err := json.NewDecoder(r).Decode(&index); err != nil {
-			return err
-		}
-
-		for i := 0; i < len(index.Manifests); i++ {
-			if index.Manifests[i].Annotations[v1.AnnotationRefName] == name {
-				d = index.Manifests[i]
-				return errEOW
-			}
-		}
-
-		return nil
-	}); err {
-	case nil:
-		return nil, fmt.Errorf("index.json: descriptor %q not found", name)
-	case errEOW:
-		return &d, nil
-	default:
-		return nil, err
+	if len(descs) == 0 {
+		return nil, fmt.Errorf("index.json: descriptor retrieved by refs %v is not match", names)
+	} else if len(descs) > 1 {
+		return nil, fmt.Errorf("index.json: descriptor retrieved by refs %v is not unique", names)
 	}
+
+	return descs, nil
 }
 
 func validateDescriptor(d *v1.Descriptor, w walker, mts []string) error {
--- a/man/oci-image-tool-create.1.md
+++ b/man/oci-image-tool-create.1.md
@@ -17,8 +17,11 @@ runtime-spec-compatible `dest/config.json`.
 **--help**
   Print usage statement
 
-**--ref**=""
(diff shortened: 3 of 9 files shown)


image/descriptor.go
```diff
@@ -66,22 +79,41 @@ func findDescriptor(w walker, name string) (*v1.Descriptor, error) {
...
+					refs = append(refs, index.Manifests[i])
+				}
+			}
+		case "platform.os":
+			for i := 0; i < len(index.Manifests); i++ {
+				if index.Manifests[i].Platform != nil && index.Manifests[i].Platform.OS == argsParts[1] {
+					refs = append(refs, index.Manifests[i])
+				}
 			}
+		case "digest":
```
wking: Walking the index to find an entry pointing at a given digest seems odd.  Doing that in `findDescriptor` seems fine, but `create` and `unpack` seem like they could skip straight over `index.json` into CAS resolution when you know the digest and media type (although you'd need to use `--type` or a new option to get the media type).

q384566678: I don't think so. when you get a given digest, we also need  `validateDescriptor` to verify that it conforms to the specification.

Mashimiao:
This PR is partly duplicated with #89

image/descriptor.go
wking: > I don't think so. when you get a given digest, we also need to verify that it conforms to the specification by `validateDescriptor`.

Sure, if you're validating a ref.  But if you're just trying to unpack an image, and you have the digest and media type (from `--type`), I don't see why you'd need to hit anything outside CAS.

q384566678:
ping @stevvooe @coolljt0725

q384566678:
@coolljt0725 @stevvooe PTAL

q384566678:

cmd/oci-image-tool/create.go
```diff
@@ -85,8 +85,8 @@ var createCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:  "ref",
-			Value: "v1.0",
-			Usage: "The ref pointing to the manifest of the OCI image. This must be present in the 'refs' subdirectory of the image.",
+			Value: "org.opencontainers.image.ref.name==v1.0",
```
stevvooe: Is this the default value? Why is the default value `v1.0`?

q384566678: > Is this the default value?

Yes.

> Why is the default value v1.0?

Just follow the previous default value.

image/descriptor.go
q384566678: What do you think about this [change](https://github.com/opencontainers/image-tools/pull/169/files#diff-da07f373046a0a24e67b18699dc15a70R177).

q384566678:
@stevvooe @xiekeyang @coolljt0725 @vbatts PTAL

q384566678:
@wking @coolljt0725 @xiekeyang @stevvooe @cyphar PTAL

q384566678:
reping @stevvooe @wking @coolljt0725 @cyphar @xiekeyang @vbatts    I need your advices.

image/descriptor.go
wking: > What do you think about this [change][1].

My problem wasn't with `validateDescriptor` for `digest==` ref names.  If we [get a media type from `--type`][2] and have a `digest==…` name, I think we can skip around [all of this][3] and never hit anything outside of CAS in `unpack`.

[1]: https://github.com/opencontainers/image-tools/pull/169/files#diff-da07f373046a0a24e67b18699dc15a70R177
[2]: https://github.com/opencontainers/image-tools/pull/169/files#r131313440
[3]: https://github.com/opencontainers/image-tools/blob/f3b94b31c161d74ce5012eb0a7cb4b9a6b95b3aa/image/image.go#L172-L179

q384566678: The type obtained by [--type](https://github.com/opencontainers/image-tools/blame/master/man/oci-image-tool-create.1.md#L27) is only the file type of `create` or `unpack` ( `imageLayout` or `image`). And the type of digest obtained here is manifest or index. We can't get the media type corresponding to digest through  `--type`, only through `findDescriptor` to confirm the media type.

q384566678:
Bump.

man/oci-image-tool-create.1.md
```diff
@@ -18,7 +18,9 @@ runtime-spec-compatible `dest/config.json`.
   Print usage statement
 
 **--ref**=""
-  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+  Specify the search criteria, format is A==B.
+  e.g. --ref org.opencontainers.image.ref.name==v1.0
+  Only support `org.opencontainers.image.ref.name`, `platform.os` and `digest` three cases.(default org.opencontainers.image.ref.name==v1.0)
```
coolljt0725: use `==` seems odd to me, can we just use `=`?  And another question, does it support the old way `--ref=latest`？ or we should use ` org.opencontainers.image.ref.name=latest` to specify a ref ?

And if we have multiple images for a `platform.os`, does all these images will be unpack?

coolljt0725: And I think in this way, we change the semantic meaning of the `ref`,  the `ref` supposed to representing a tag of image.

q384566678: >use == seems odd to me, can we just use =? 

updated.
>And another question, does it support the old way --ref=latest？ or we should use org.opencontainers.image.ref.name=latest to specify a ref ?

use org.opencontainers.image.ref.name=lates

>And if we have multiple images for a platform.os, does all these images will be unpack?

Yes, all images conforming to the platform. OS will be unpack.

coolljt0725: > Yes, all images conforming to the platform. OS will be unpack.

Is that a desired behavior ?  Will these images be unpack to different `rootfs` ?

q384566678: I think maybe something is not clear:
unpack at a time can only be an `image` or `imageLayout` unpack. When `ref` point to the `index`, then go to the `platform.os` to filter manifest, and then according to the manifests to unpack to the `rootfs`, I think your question is not that if the match of the manifest there are many, should not unpack to the same `rootfs`?
If this is the problem, I do not know is not able to unpack to the same file, if not, it should increase the screening of the manifests to determine.

coolljt0725: There could be multiple images with different ref for a specific platform in an `image` or `imageLayer`, My question is if `platform.os=xx` match multiple  image,  when you do `unpack` or `create bundle`, Will these images will be unpacked to same `rootfs`? That's not a desired behavior.

q384566678: Sorry. I did not understand it before.

>My question is if platform.os=xx match multiple image, when you do unpack or create bundle, Will these images will be unpacked to same rootfs?

No. if platform.os=xx match multiple image, The program will be [error](https://github.com/opencontainers/image-tools/pull/169/files#diff-4d7e922f1104313b276b759505857ed7R113).

q384566678:
ping @wking @xiekeyang @stevvooe @vbatts @cyphar

coolljt0725:
I just think `org.opencontainers.image.ref.name` is tool long for user to type it, can we make a short name like `--ref name=xx`? or just keep the old way `--ref xxx` to stand for `--ref org.opencontainers.image.ref.name=xxx`?

q384566678:
@coolljt0725 upadted. Thanks for your advices.

coolljt0725:
with this patch when I validated a image, the output is 
```
$ oci-image-tool validate busybox/
oci-image-tool: reference "linux": OK
oci-image-tool: reference "sha256:40a114053d955a2b80ee2cf6e13410b28b59594ceee9036b41e12c42d3e16615": OK
busybox/: OK
Validation succeeded

```
I think it's a bit of redundant and make user confuse,

and I use `oci-image-tool validate --ref name=latest busybox/` , it doesn't work, 
```
oci-image-tool validate  --ref name=latest busybox/
1 errors detected:
busybox/: validation failed: reference name=latest not found

```
So the `oci-image-tool validate` has a different UI for `--ref`? It's better to keep consistent.

q384566678:
Modified to become the following result, and I think this will be better.
```
 ./oci-image-tool validate ../ubuntu
oci-image-tool: reference "name=latest": OK
oci-image-tool: reference "platform.os=linux": OK
oci-image-tool: reference "digest=sha256:ab8bbdc63526cb49eba4a9dc6833bc59ae978c0d0234726939ea762a26beb396": OK
../ubuntu: OK
Validation succeeded

./oci-image-tool validate --ref name=latest ../ubuntu
oci-image-tool: reference "name=latest": OK
../ubuntu: OK
Validation succeeded

**Comment by q384566678** [comment]
@coolljt0725 @wking @stevvooe @xiekeyang @vbatts @cyphar Any other comments?

**Comment by q384566678** [comment]

**Comment by q384566678** [comment]
@coolljt0725 @xiekeyang PTAL

**Comment by q384566678** [comment]
ping @coolljt0725 @xiekeyang

**Comment by coolljt0725** [comment]
I just think the output is a bit of redundant if we don't specify any ref.
```
$ oci-image-tool validate busybox/
oci-image-tool: reference "linux": OK
oci-image-tool: reference "sha256:40a114053d955a2b80ee2cf6e13410b28b59594ceee9036b41e12c42d3e16615": OK
busybox/: OK
Validation succeeded
```
All these three `OK` are for a same image, but this output  make users think there are three images. Can we only display a `OK` if we don't specify a ref ?Or make the output more user friendly?

File: man/oci-image-tool-create.1.md [hunk 4]
```diff
@@ -18,7 +18,9 @@ runtime-spec-compatible `dest/config.json`.
   Print usage statement
 
 **--ref**=""
-  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+  Specify the search criteria, format is A=B.
+  e.g. --ref name=v1.0
+  Only support `name`, `platform.os` and `digest` three cases.(default org.opencontainers.image.ref.name=v1.0)
```
coolljt0725: `(default org.opencontainers.image.ref.name=v1.0)` needs update

man/oci-image-tool-unpack.1.md
```diff
@@ -15,7 +15,9 @@ oci-image-tool unpack \- Unpack an image or image source layout
   Print usage statement
 
 **--ref**=""
-  The ref pointing to the manifest to be unpacked. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+  Specify the search criteria, format is A=B.
+  e.g. --name=v1.0
+  Only support `name`, `platform.os` and `digest` three cases.(default org.opencontainers.image.ref.name=v1.0)
```
coolljt0725: same here

man/oci-image-tool-validate.1.md
```diff
@@ -19,6 +19,7 @@ oci-image-tool validate \- Validate one or more image files
   The reference to validate (should point to a manifest).
   Can be specified multiple times to validate multiple references.
   `NAME` must be present in the `refs` subdirectory of the image.
+  Format is A=B. Only support 'name', 'platform.os' and 'digest' three cases.
```
coolljt0725: Does it still work if user specify multiple ref? I haven't tested this, please test this.

q384566678:
```
➜  ~ oci-image-tool validate --ref platform.os=linux --ref name=latest busybox
oci-image-tool: reference "platform.os=linux": OK
oci-image-tool: reference "name=latest": OK
busybox: OK
Validation succeeded
```

It's ok.

q384566678:
@coolljt0725 updated, PTAL.

q384566678:
ping @jonboulle

q384566678:
ping @coolljt0725 @xiekeyang

q384566678:

q384566678:
@coolljt0725 @stevvooe @xiekeyang @cyphar @vbatts PTAL

q384566678:

q384566678:
@opencontainers/image-tools-maintainers @vbatts @coolljt0725 @stevvooe @Mashimiao 
I updated this Pr, and now I change the following:

1. Validate changes to validate a file once
2. Find reference more accurately through multiple ref inputs

The following is the specific operation results:
```
oci-image-tool validate --ref name=latest --ref platform.os=linux ubuntu.tar 
oci-image-tool: reference [name=latest platform.os=linux]: OK
ubuntu.tar: OK
Validation succeeded
```

image/descriptor.go
```diff
@@ -66,22 +84,49 @@ func findDescriptor(w walker, name string) (*v1.Descriptor, error) {
 			return err
 		}
 
-		for i := 0; i < len(index.Manifests); i++ {
-			if index.Manifests[i].Annotations[v1.AnnotationRefName] == name {
-				d = index.Manifests[i]
-				return errEOW
+		descs = index.Manifests
+		for _, name := range names {
```
Mashimiao: The following code is totally duplicate with part code of validate() in image/image.go
Maybe we can create a common or utility function by taking advantage of it.

q384566678: updated, PTAL.

Mashimiao:
LGTM

coolljt0725:
ping @opencontainers/image-tools-maintainers  PTAL. This is a change on UI, I think it needs more lgtm

stevvooe:
It would be good to summarize the change. Does this now allow one to ignore parts of the layout by specifying a ref? What is the use for only validating one part of an image?

Is there an updated proposal?

q384566678:
> Does this now allow one to ignore parts of the layout by specifying a ref?

I don't quite understand what you mean. What does `parts of the layout` mean? 
I think the only thing that's been ignored is the manifest that doesn't match the lookup information when the validation is performed.

> What is the use for only validating one part of an image?

For the image validation process, I just continued the previous practice: Verify the specified manifest according to the ref.


In my opinion, this patch is no problem and can be merged.

q384566678:
This patch dragged for a long time, I will merge it first, what problems can be amended later.

stevvooe:
@q384566678 Did this patch have the requisite LGTMs from maintainers to proceed? Questions in https://github.com/opencontainers/image-tools/pull/169#issuecomment-354897427 were not resolved.

q384566678:
>Did this patch have the requisite LGTMs from maintainers to proceed? 

According to the latest rules, a LGTM can merge patches. This patch waits for a long time, and no one else to review, so I can only merge it first.

>Questions in #169 (comment) were not resolved.

As for the question you said, I have already given a reply below because I did not understand something you said and you did not reply to me. I think if there are other problems can be repaired in the next pr.

Please forgive me if there is anything wrong with it.

stevvooe:
I reviewed it and asked for a summary of the changes: "It would be good to summarize the change." Just because it takes a long time doesn't mean it is okay to merge.

I have no clue what is wrong with this because there is no proposal describing the end result of what this PR is attempting to accomplish.

q384566678:
The purpose of this pr is to solve the problem that the `descriptor` is not unique when searching `descriptor` by the input conditions. After the change can enter multiple conditions to search, to ensure that the search `descriptor` is unique.
So what do you think there is something to change?

PR #170, opened 2017-08-02
image.go: Simplified type declaration

q384566678:


q384566678:
ping @coolljt0725 @xiekeyang @stevvooe PTAL

coolljt0725:
LGTM

stevvooe:
LGTM

Caveat: let's question the fact that we have 4 string arguments in a row. This could be a bug waiting to happen.

PR #172, opened 2017-08-04
cleanup created bundle when creating failed

Mashimiao:
If any error occurs, we should cleanup the bundle we created. If not, this may confuse users.

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -290,7 +290,7 @@ func createRuntimeBundle(w walker, dest, refName, rootfs, platform string) error
 	return nil
 }
 
-func createBundle(w walker, m *manifest, dest, rootfs string) error {
+func createBundle(w walker, m *manifest, dest, rootfs string) (retErr error) {
 	c, err := findConfig(w, &m.Config)
 	if err != nil {
 		return err
@@ -301,6 +301,13 @@ func createBundle(w walker, m *manifest, dest, rootfs string) error {
 			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
 				return err2
 			}
+			defer func() {
+				if retErr != nil {
+					if err3 := os.RemoveAll(dest); err3 != nil {
+						fmt.Printf("Failed to clean up %q: %s\n", dest, err3.Error())
+					}
+				}
+			}()
 		} else {
 			return err
 		}


image/image.go
```diff
@@ -301,6 +301,13 @@ func createBundle(w walker, m *manifest, dest, rootfs string) error {
 			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
 				return err2
 			}
+			defer func() {
+				if err != nil {
+					if err3 := os.RemoveAll(dest); err3 != nil {
+						fmt.Printf("warning: failed to clean up %q\n", dest)
```
wking: Can we wrap `err3`?  I'd like to expose the reason for the failure to the user.

Mashimiao: I think we can't, because we can't return any value in defer function.

wking: Can you squash it into `err`?  Or at least print it here?

Mashimiao: Let's print it here, updated

wking: > Let's print it here, updated

Did you forget to push?  I still see this PR pointing at c5ed3a8ec, has a 2017-08-04 commit date.

Mashimiao: Oh,  my fault.  updated.

image/image.go
```diff
@@ -301,6 +301,13 @@ func createBundle(w walker, m *manifest, dest, rootfs string) error {
 			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
 				return err2
 			}
+			defer func() {
+				if err != nil {
+					if err3 := os.RemoveAll(dest); err3 != nil {
```
coolljt0725: With this defer, it may be good to use named returns.

Mashimiao: Sorry, I don't know what you mean. What kind of named return?

coolljt0725:
```
func createBundle(w walker, m *manifest, dest, rootfs string) ( retEr error)
```
and use 
```
defer func() {
     if retErr != nil {
}
}()
```
here

Mashimiao: It seems still can't return err and err3 at same time.

coolljt0725: why `err3` need to return?

coolljt0725: see https://github.com/opencontainers/image-tools/blob/master/image/manifest.go#L109

Mashimiao: Fine, updated

image/image.go
```diff
@@ -301,6 +301,13 @@ func createBundle(w walker, m *manifest, dest, rootfs string) error {
 			if err2 := os.MkdirAll(dest, 0755); err2 != nil {
 				return err2
 			}
+			defer func() {
+				if retErr != nil {
```
wking: Can we use `err` instead of `retErr`.  Otherwise I expect this will always be `nil` (unless we change to set `retErr` in the body of `createBundle`).  Or maybe Go does something magic with named returns to set them up before defers fire?

coolljt0725: when return ,go will assign the return value to `retErr`

coolljt0725:
LGTM

q384566678:
LGTM

PR #173, opened 2017-08-21
validate: Increase ref content repeat detection

q384566678:
When the input ref content is repeated, should output a prompt.


q384566678:
@coolljt0725 PTAL

coolljt0725:
LGTM

PR #177, opened 2017-08-30
doc.go: Perfect description

q384566678:

### DIFF
--- a/image/doc.go
+++ b/image/doc.go
@@ -12,5 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package image defines methods for validating, and unpacking OCI images.
+// Package image defines methods for validating, unpacking OCI images and creating OCI runtime bundle.
 package image


image/doc.go
```diff
@@ -12,5 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package image defines methods for validating, and unpacking OCI images.
+// Package image defines methods for validating, unpacking and creating OCI images.
```
wking: What functionality does it provide for creating images?  The only creation functionality I can find is for bundle creation (e.g. [`CreateRuntimeBundle`][1]), and that's part of unpacking an image.

I'd expect image creation to include layer generation (in flight with #8), CAS pushing (in flight with #5), and/or index generation (something like it in #5, but I'm not aware of any image-tools work on this front since opencontainers/image-spec#533).

[1]: https://github.com/opencontainers/image-tools/blob/cfb23a2546eeef2f9767bd54c316fc3ed8686480/image/image.go#L236-L241

q384566678: > What functionality does it provide for creating images? 

createBundle

q384566678: I think the new changes can express my thoughts.

image/doc.go
```diff
@@ -12,5 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package image defines methods for validating, and unpacking OCI images.
+// Package image defines methods for validating, unpacking OCI images and creating runtime bundle.
```
Mashimiao: `OCI runtime bundle` will be better

q384566678: updated.

Mashimiao:
LGTM

stevvooe:
LGTM

PR #178, opened 2017-08-31
validate.go: Modify the output mode

q384566678:


Mashimiao:
There is no different between new and old code.

q384566678:
I'm trying to keep it uniform, and I wonder if I can delete `stdout` and turn it into `fmt`. Because it was changed in both `create` and `unpack`.

PR #180, opened 2017-09-04
image: add a warning when the platform is not applicable

q384566678:


image/image.go
```diff
@@ -179,6 +179,10 @@ func unpack(w walker, dest, refName, platform string) error {
 	}
 
 	if ref.MediaType == validRefMediaTypes[0] {
+		if platform != "" {
+			fmt.Println("WARNING: reftype manifest does not support platform option, which are only applicable if reftype is index.")
```
Mashimiao: In `--platform`'s help message, we already have such hint, I think there it no need to notify user again

PR #182, opened 2017-09-07
Rename `oci-image-tool create` to `oci-image-tool create-bundle`

coolljt0725:
`oci-image-tool create` is confusing, rename it to `create-bundle`,
and we may have `create-layer` in future which infligh in


wking:
On Thu, Sep 07, 2017 at 08:27:01AM +0000, Lei Jitang wrote:
> `oci-image-tool create` is confusing, rename it to `create-bundle`,
> and we may have `create-layer` in future which infligh in #8.

opencontainers/image-spec#467 is still open and
opencontainers/image-spec#554 was rejected, but I still prefer
‘unpack’ for this sort of thing [1].  That focuses the name on the
image-spec object (the manifest and blobs) that are being unpacked,
instead of on the runtime-spec object (the bundle) that is being
created.  To distinguish from other unpacking (e.g. unpacking a single
layer, or a stack of layers, just a config, etc.), I'd recommend
‘unpack manifest’ (or if you prefer, ‘unpack-manifest’).  Unpacking a
single layer would be ‘unpack layer’, unpacking multiple layers would
be ‘unpack layers’ and unpacking an image-spec config into a
runtime-spec config would be ‘unpack config’.

[1]: https://github.com/wking/image-spec/blob/cb65bdf4c29e7967b0ebe876809c8292d3bd3628/spec.md#actions

cmd/oci-image-tool/bundle.go
```diff
@@ -76,7 +76,7 @@ func createHandle(context *cli.Context) error {
 }
 
 var createCommand = cli.Command{
-	Name:   "create",
+	Name:   "create-bundle",
 	Usage:  "Create an OCI image runtime bundle",
```
q384566678: Is it better to change this file name to `create-bundle.go`?

jonboulle:
I tend to agree with https://github.com/opencontainers/image-tools/pull/182#issuecomment-328409898 - unpack nomenclature seems a lot more intuitive here (especially considering this is a superset of the `unpack` command)

coolljt0725:
@jonboulle  @wking  we already has a `oci-image-tool unpack` which only unpack the layers to a rootfs, so do you mean to use `oci-image-tool unpack` to create runtime bundle and use `oci-image-tool unpack-layers` to unpack the layers to a rootfs?

wking:
On Thu, Oct 19, 2017 at 01:17:08AM +0000, Lei Jitang wrote:
> … we already has a `oci-image-tool unpack` which only unpack the
> layers to a rootfs, so do you mean to use `oci-image-tool unpack` to
> create runtime bundle and use `oci-image-tool unpack-layers` to
> unpack the layers to a rootfs?

I'd rather avoid an unqualified ‘unpack’.  I don't mind about
hyphenated or unhyphenated versions, but here's the unhyphenated forms
of what I'm thinking (based on my earlier suggestions in [1]):

* ‘unpack layers’ unpacks a stack of layers (as referenced by a
  manifest).  Currently this is ‘unpack’.

* ‘unpack layer’ unpacks a single layer (possibly into an existing,
  populated directory).  We don't have a command for this at the
  moment.

* ‘translate config’ (or ‘unpack config’) for converting a
  configuration from the image-spec form to the runtime-spec form (and
  possibly vice versa via --from and --to options) and writing the
  result to stdout.

* ‘unpack manifest’ combining ‘unpack layers’ and ‘translate config’
  with the config written to config.json.  Currently this is ‘create’.

[1]: https://github.com/opencontainers/image-tools/pull/182#issuecomment-328409898

PR #184, opened 2017-09-09
validate: remove autodetect for type and add require limit

Mashimiao:
1. spec says any extra fields in json file can be ignored, so I think
we can't clearly distingish which type the text file is, we'd better remove
autodetect for text files.
2. without audotdetect, we must requrie to user to set file type for validation

### DIFF
--- a/man/oci-image-tool-validate.1.md
+++ b/man/oci-image-tool-validate.1.md
@@ -20,15 +20,15 @@ oci-image-tool validate \- Validate one or more image files
   Reference should point to a manifest or index.
   e.g. --ref name=v1.0 --ref platform.os=latest
   Only support `name`, `platform.os` and `digest` three cases.
-  Only applicable if type is image or imageLayout.
+  Only applicable if type is image.
 
 **--type**=""
-  Type of the file to validate. If unset, oci-image-tool will try to auto-detect the type. One of "imageLayout,image,imageZip,manifest,imageIndex,config"
+  Type of the file to validate. One of "image,manifest,imageIndex,config"
 
 # EXAMPLES
 ```
 $ skopeo copy docker://busybox oci:busybox-oci:latest
-$ oci-image-tool validate --type imageLayout --ref name=latest busybox-oci
+$ oci-image-tool validate --type image --ref name=latest busybox-oci
 busybox-oci: OK
 ```
 
--- a/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
+++ b/cmd/oci-create-runtime-bundle/oci-create-runtime-bundle.1.md
@@ -1,0 +1,46 @@
+% OCI(1) OCI-CREATE-RUNTIME-BUNDLE User Manuals
+% OCI Community
+% JULY 2016
+# NAME
+oci-create-runtime-bundle \- Create an OCI runtime bundle
+
+# SYNOPSIS
+**oci-create-runtime-bundle** [src] [dest] [flags]
+**oci-create-runtime-bundle** [--help|-v|--version]
+
+# DESCRIPTION
+`oci-create-runtime-bundle` validates an application/vnd.oci.image.manifest.v1+json and unpacks its layered filesystem to `dest/rootfs`, although the target directory is configurable with `--rootfs`. See **oci-unpack**(1) for more details on this process.
+
+Also translates the referenced config from application/vnd.oci.image.config.v1+json to a
+runtime-spec-compatible `dest/config.json`.
+
+# FLAGS
+**--help**
+  Print usage statement
+
+**--ref**=""
+  The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image. (default "v1.0")
+
+**--rootfs**=""
+  A directory representing the root filesystem of the container in the OCI runtime bundle. It is strongly recommended to keep the default value. (default "rootfs")
+
+**--type**=""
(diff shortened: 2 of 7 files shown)


Mashimiao:

Mashimiao:
ping @coolljt0725  @xiekeyang  @q384566678  @stevvooe

q384566678:
I think in `create` and `unpack` also need to make the appropriate changes.

Mashimiao:
create and unpack doesn't have text file type autodetect problem.  Currently, whether type is required or not does not break their function. Let's solve validate's problem first. Then if necessary, we can make unified changes in another PR.

q384566678:
LGTM

This can also solve the problem in #62 . Wait for other's opinion.

man/oci-image-tool-validate.1.md
```diff
@@ -22,7 +22,7 @@ oci-image-tool validate \- Validate one or more image files
   Only applicable if type is image or imageLayout.
 
 **--type**=""
-  Type of the file to validate. If unset, oci-image-tool will try to auto-detect the type. One of "imageLayout,image,imageZip,manifest,imageIndex,config"
+  Type of the file to validate. One of "imageLayout,image,manifest,imageIndex,config"
```
coolljt0725: Can we set a default type if `--type` is unset? It seems odd if it doesn't work without any options.

Mashimiao: I don't think it's a good idea. Because `--type` is a little special, it's a must be required option but there is not a value could be used for all files, that means it's almost impossible to find a suitable default value.
So, I think let it to be unset will be better.

coolljt0725: >   there is not a value could be used for all files, that means it's almost impossible to find a suitable default value.

Default is not a value for all files, it's value are most common.  From that I know, it's not a good design from user perspective if a cli requires a `option` .  


And I tested the patch
```
lei@ubuntu:~/images$ oci-image-tool validate busybox/
--type must be set
lei@ubuntu:~/images$ oci-image-tool validate --type imageLayout busybox/
1 errors detected:
busybox/: type "imageLayout" unimplemented
lei@ubuntu:~/images$ oci-image-tool validate --type image busybox/
autodetected image file type is: imageLayout
busybox/: OK
Validation succeeded

```
Is the output `autodetected image file type is: imageLayout` necessary and why  use ` --type imageLayout` doesn't work but the autodetected image type is `imageLayout`?`

Mashimiao: Oh, sorry....  that's my fault. I should remove imageLayout from help and manpage.
In my design, users can only specify `image,manifest,imageIndex,config` to `--type`. 
If we get that users want to validate an image, then we try to detect its type, may be a layout, tar file or zip file.
As you mentioned at https://github.com/opencontainers/image-tools/pull/141#issuecomment-326907714, I try to use imageType to handle all image format.

Mashimiao: Updated, removed imageLayout. PTAL

q384566678:
Also need to fix `completions/bash/oci-image-tool`.

Mashimiao:
updated

q384566678:
In general I agree this PR. @coolljt0725, WDYT?

Mashimiao:

q384566678:
LGTM

ping @coolljt0725 @xiekeyang PTAL

q384566678:
ping @coolljt0725 @xiekeyang

Mashimiao:
ping @coolljt0725

coolljt0725:
ping @opencontainers/image-tools-maintainers   @wking  PTAL.  I'm not quite sure about this.

q384566678:

Mashimiao:

vbatts:
@Mashimiao is the autodetect causing issues? I'm am all for reducing magic-detection, but if this is not causing issues, then what is the issue?

Mashimiao:
Autodetect really causes issue. But I think that happens just when we use it for text file detection.
If we specified to validate the whole image not image files,  it can easily and clearly find out the type of image. If the image is a directory, it must be a imageLayout. If not a directory, we can find out it content type and call validation function for that type.
So, I just removed the autodetec for image files(like index, config files), but retain the autodetect for image type.

Mashimiao:
ping @vbatts

man/oci-image-tool-validate.1.md
```diff
@@ -19,15 +19,15 @@ oci-image-tool validate \- Validate one or more image files
   The reference to validate (should point to a manifest).
   Can be specified multiple times to validate multiple references.
   `NAME` must be present in the `refs` subdirectory of the image.
-  Only applicable if type is image or imageLayout.
+  Only applicable if type is image.
 
 **--type**=""
-  Type of the file to validate. If unset, oci-image-tool will try to auto-detect the type. One of "imageLayout,image,imageZip,manifest,imageIndex,config"
+  Type of the file to validate. One of "image,manifest,imageIndex,config"
```
stevvooe: An `image` isn't the layout. The layout is the collection of resources that can make up an image. For example, I can still have an image and not have a layout.

Mashimiao: Yeah, agree.  And as you said, layout makes up an image. If we have a layout then we have an image. The point is image may be stored in different file types.
But if you don't have a layout, then it will not be a valid OCI image. 
So, how do you think we should change the help info?

q384566678:
LGTM

PR #190, opened 2017-10-18
Optimize output

q384566678:

### DIFF
--- a/cmd/oci-image-tool/create.go
+++ b/cmd/oci-image-tool/create.go
@@ -72,7 +72,12 @@ func createHandle(context *cli.Context) error {
 
 	}
 
-	return err
+	if err != nil {
+		return fmt.Errorf("Creating failed: %v", err)
+	}
+
+	fmt.Println("Creating succeeded")
+	return nil
 }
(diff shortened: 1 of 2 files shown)


cmd/oci-image-tool/create.go
```diff
@@ -72,7 +72,12 @@ func createHandle(context *cli.Context) error {
 
 	}
 
-	return err
+	if err != nil {
+		return fmt.Errorf("Creating failed: %v", err)
```
jonboulle: should probably use pkg/errors

cmd/oci-image-tool/create.go
```diff
@@ -72,7 +72,12 @@ func createHandle(context *cli.Context) error {
...
+	}
+
+	fmt.Println("Creating succeeded")
```
jonboulle: if you really want this output, it should go to stderr, not stdout

Mashimiao:
LGTM

stevvooe:
What is the purpose of this change? What do you mean by "Optimize Output"?

PR #192, opened 2017-10-19
191 docs fix skopeo copy cmd

david-drinn:
The change is simple, but I re-did the commit message to better follow the contributing guidelines, but unfortunately not before I had already published it -- therefore three commits.


david-drinn:
Redo.

Issue #166, opened 2017-07-26
image: assumes that container has sh in PATH


stevvooe:
The current conversion code makes the assumption that `sh` is available when `Process.Args` is empty. This isn't covered in the specification and `sh` may not be available within the container, leading to confusing errors.

PR #196, opened 2018-01-07
Remove Default sh in Process Args

bkochendorfer:
Defaulting to sh is not covered in the spec.

Signed-off-by: Brett Kochendorfer <brett.kochendorfer@gmail.com>

Closes https://github.com/opencontainers/image-tools/issues/166


bkochendorfer:
Was this the right approach or not what you were looking for with regards to the issue?

q384566678:
`Args` in the `spec` have the following provisions:

>This specification extends the IEEE standard in that at least one entry is REQUIRED,

So I think `args` should not be empty, if `sh` is wrong, Is it possible to replace `sh` with a correct value?

bkochendorfer:
@q384566678 That's a good question. Would it be better here to complain that no arg was provided rather than defaulting to `sh`?

PR #199, opened 2018-02-05
Revert "Relax LGTMs"

stevvooe:
This reverts commit e073febd8714374d6b459924cb0e2f2f998f6252.

This could only change with a proper vote by the project maintainers.
Pushing this through with only 2 maintainers was not acceptable.

Furthermore, if there are problems getting PRs through, it usually has
nothing to do with the number of maintainers on the project. It
typically comes down to the level of engineering work put into making
sure that a change is a clear and accurate solution to the problem. If
it is unclear whether or not a change has impact, then the default
reaction is to not merge the PR.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -5,7 +5,7 @@ requirements:
     required: true
 
 group_defaults:
-  required: 1
+  required: 2
   approve_by_comment:
     enabled: true
     approve_regex: '^(Approved|lgtm|LGTM|:shipit:|:star:|:\+1:|:ship:)'


coolljt0725:
I'm agree with this revert. LGTM

FlorianLudwig:
> Furthermore, if there are problems getting PRs through, it usually has
nothing to do with the number of maintainers on the project. It
typically comes down to the level of engineering work put into making
sure that a change is a clear and accurate solution to the problem. If
it is unclear whether or not a change has impact, then the default
reaction is to not merge the PR.

@stevvooe does that explain why this one never got merged?

cyphar:
LGTM.

cyphar:
@FlorianLudwig This project is no longer actively maintained, which is the core issue. Steven is right that in an actively-maintained project with multiple active maintainers, the 2-LGTM rule is rarely an issue. However, the project has the same maintenance issues as it did when the 1-LGTM change was merged, sort of proving Steven's point.

PR #206, opened 2018-06-12
image: fix the interaction of findManifest and findConfig with walker.find

### LINKED ISSUE #205: oci-image-tool commands ‘validate’ and ‘create’ failing for tar and zip images
Due to a change in rev. 9027d58 validating an image in a tar (or zip) file is failing currently with the following message:

    <tar-image>: find failed: unable to walk: end of walk

This is because in the context of the transition from using `w.walk` to `w.find` in `findManifest` (and, similar, `findConfig`) which has been done in 9027d58, the interaction between `walk/findFunc` and error handling, especially regarding errEOW, appears not to have been adjusted accordingly.

I have created a patch https://github.com/knieriem/image-tools/commit/06a4e9658be3d997a601d6b13987459570227901 that describes in more detail what is happening, and adjusts how `image.findManifest` and `image.findConfig` interact with `image.walker.find` so that `validate` and `create` commands work again for _tar_ and _zip_ files.

knieriem:
Before rev. 9027d58 findManifest and findConfig called w.walk directly,
providing a walkFunc that returned errEOW in case a file has been
found and processed successfully.  A switch construct evaluating the
error-result of w.walk then tested for the errEOW case, meaning the file
has been found.

Starting with rev. 9027d58 the call to w.walk in findManifest and
findConfig has been replaced by w.find. Though, the function literal
provided by findManifest and findConfig to w.find -- now a findFunc,
not a walkFunc anymore -- still returns errEOW. As a consequence,
walker.find implementations for tar and zip would treat errEOW as an
error, making oci-image-tool 'validate' and 'create' stop with an error:

	<tar-image>: find failed: unable to walk: end of walk

This patch adjusts tarWalker's and zipWalker's implementations of
walker.find, using errEOW to stop w.walk early in case of success; there
is no need for the 'done' variable anymore. Additionally, places where
w.find is called, functions findManifest, unpackManifest, and findConfig,
have been adapted so that their findFunc argument to w.find does not
return errEOW anymore; instead nil is returned on success. Consequently,
the switch constructs now test err for 'nil' instead of 'errEOW', and
'os.ErrNotExist' instead of nil.

Fixes #205

Signed-off-by: Michael Teichgräber <mteichgraeber@gmx.de>

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -48,11 +49,11 @@ func findConfig(w walker, d *v1.Descriptor) (*v1.Image, error) {
 			return err
 		}
 
-		return errEOW
+		return nil
 	}); err {
-	case nil:
+	case os.ErrNotExist:
 		return nil, fmt.Errorf("%s: config not found", cpath)
-	case errEOW:
+	case nil:
 		return &c, nil
 	default:
 		return nil, err
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -53,11 +53,11 @@ func findManifest(w walker, d *v1.Descriptor) (*v1.Manifest, error) {
 			return err
 		}
 
-		return errEOW
+		return nil
 	}); err {
-	case nil:
+	case os.ErrNotExist:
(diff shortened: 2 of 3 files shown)


edtan:
Just wondering if there's anything blocking this PR?  I ran into this issue with some `buildkit`-generated `tar`s when trying to use the `validate` and `create` commands.  (I was using this tool to learn more about the OCI specs.)  Thanks!

PR #211, opened 2019-02-26
updating CoC link and adding security section

vsoch:
Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

https://github.com/opencontainers/org/issues/3

PR #212, opened 2019-02-27
updating code of conduct and security links

vsoch:
Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

This pull request will replace #211 with a better commit message, single commit. It looks like the copy paste from the file resulted in different characters for the file tree, which I don't think should be an issue (but let me know if you have concerns).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,23 +60,26 @@ $ oci-image-tool unpack --ref latest busybox-oci busybox-bundle
 $ tree busybox-bundle
 busybox-bundle
 ├── bin
-│   ├── [
-│   ├── [[
-│   ├── acpid
-│   ├── addgroup
-│   ├── add-shell
+│   ├── [
+│   ├── [[
+│   ├── acpid
+│   ├── addgroup
+│   ├── add-shell
 [...]
 ```
 
 # Contributing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,5 +177,7 @@ Read more on [How to Write a Git Commit Message](http://chris.beams.io/posts/git
 8. When possible, one keyword to scope the change in the subject (i.e. "README: ...", "runtime: ...")
 
 
+[security]: https://github.com/opencontainers/org/blob/master/security
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/.github/CODE_OF_CONDUCT.md
```
coolljt0725: The link should be 
`https://github.com/opencontainers/org/blob/master/code-of-conduct.md` and  `https://github.com/opencontainers/org/blob/master/security.md`

vsoch:
There is a PR under review that changes the security markdown file to a folder, so the link is correct. We can double check here when it’s merged before merging here if you like!

vsoch:
Code of conduct link is updated to be outside of .github (and in all caps) to reflect discussion in OCI call. @coolljt0725 you should be able to check the links now (the change is merged).

PR #215, opened 2019-03-06
update to golang 1.11

q384566678:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - 1.7.x
-  - 1.8.x
-  - 1.9.x
+  - 1.10.x
+  - 1.11.x
 
 sudo: required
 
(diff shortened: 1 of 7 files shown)


q384566678:

xiekeyang:
@q384566678 
commit 4775f434bc5a5f1a55505b2adddf0101d3dc46b2 might better to be another PR, right?
To `update to golang 1.11` one, I'm OK.

q384566678:
@xiekeyang The previous mistakes were all before go1.10, and I don't think there is any change. So I think it is better to have two commits in the same pr.

q384566678:
I added some instructions to the commit.　@xiekeyang PTAL

xiekeyang:
Got it!

LGTM

PR #217, opened 2019-03-20
Fix ref option in README

ArkArk:
In `README.md`, I think the following command is correct:
```sh
$ oci-image-tool create --ref name=latest busybox-oci busybox-bundle
```


ArkArk:
I'm sorry that this PR is duplicated with https://github.com/opencontainers/image-tools/pull/207

PR #224, opened 2020-06-22
README: add summary of project status

cyphar:
We should forward users to umoci (which is now part of the OCI), because
of the many issues with oci-image-tools in the area of image
manipulation. Sadly there is no obvious home to move the validation
code, so we can just tell users to be aware that the validation code is
not being actively maintained.

/cc @opencontainers/image-tools-maintainers
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,42 @@
 `oci-image-tool` is a collection of tools for working with the [OCI image format specification](https://github.com/opencontainers/image-spec).
 To build from source code, image-tools requires Go 1.7.x or above.
 
+## Project Status
+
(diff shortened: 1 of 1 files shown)


cyphar:
@opencontainers/image-tools-maintainers Can we please get this merged so people aren't confused about what tools the OCI recommends using?

vbatts (approved):
LGTM

cyphar:
LGTM -- let me see if I can self-LGTM...

coolljt0725 (approved):
LGTM

